PR #2, opened 2020-05-21
Require that test targets have a host application

NickEntin:
Without a host application, many of the generated accessibility properties in UIKit won't be populated correctly, which can lead to things silently failing in less-than-obvious ways. This adds an assertion at the beginning of the test methods to check that the host application exists, and adds a note to the README about this requirement.

### DIFF
--- a/AccessibilitySnapshot/Classes/FBSnapshotTestCase+Accessibility.swift
+++ b/AccessibilitySnapshot/Classes/FBSnapshotTestCase+Accessibility.swift
@@ -44,11 +44,7 @@ extension FBSnapshotTestCase {
         file: StaticString = #file,
         line: UInt = #line
     ) {
-        // The tests must be run in a host application in order for the accessibility properties to be populated
-        // correctly. The `UIApplication.shared` singleton is non-optional, but will be uninitialized when the tests are
-        // running outside of a host application, so we can use this check to determine whether we have a test host.
-        let hostApplication: UIApplication? = UIApplication.shared
-        guard hostApplication != nil else {
+        guard isRunningInHostApplication else {
             XCTFail(
                 "Accessibility snapshot tests cannot be run in a test target without a host application",
                 file: file,
@@ -180,6 +176,16 @@ extension FBSnapshotTestCase {
         if requiresWindow {
             view.removeFromSuperview()
         }
+    }
+
+    // MARK: - Internal Properties
+
+    var isRunningInHostApplication: Bool {
+        // The tests must be run in a host application in order for the accessibility properties to be populated
+        // correctly. The `UIApplication.shared` singleton is non-optional, but will be uninitialized when the tests are
+        // running outside of a host application, so we can use this check to determine whether we have a test host.
+        let hostApplication: UIApplication? = UIApplication.shared
+        return (hostApplication != nil)
     }
 
 }
--- a/AccessibilitySnapshot/Classes/FBSnapshotTestCase+Accessibility.swift
+++ b/AccessibilitySnapshot/Classes/FBSnapshotTestCase+Accessibility.swift
@@ -44,6 +44,19 @@ extension FBSnapshotTestCase {
         file: StaticString = #file,
         line: UInt = #line
     ) {
+        // The tests must be run in a host application in order for the accessibility properties to be populated
+        // correctly. The `UIApplication.shared` singleton is non-optional, but will be uninitialized when the tests are
+        // running outside of a host application, so we can use this check to determine whether we have a test host.
+        let hostApplication: UIApplication? = UIApplication.shared
+        guard hostApplication != nil else {
+            XCTFail(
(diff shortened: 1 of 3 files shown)


AccessibilitySnapshot/Classes/FBSnapshotTestCase+Accessibility.swift
```diff
@@ -44,6 +44,19 @@ extension FBSnapshotTestCase {
...
     ) {
+        // The tests must be run in a host application in order for the accessibility properties to be populated
+        // correctly. The `UIApplication.shared` singleton is non-optional, but will be uninitialized when the tests are
+        // running outside of a host application, so we can use this check to determine whether we have a test host.
+        let hostApplication: UIApplication? = UIApplication.shared
+        guard hostApplication != nil else {
+            XCTFail(
+                "Accessibility snapshot tests cannot be run in a test target without a host application",
+                file: file,
+                line: line
```
dfed, line 55: Something of a bit, but I’d consider originating this failure from your code, rather than from the calling file/line. This error isn’t caused by the call site but rather the configuration of the test.

AccessibilitySnapshot/Classes/FBSnapshotTestCase+ObjCSupport.swift
```diff
@@ -49,6 +49,14 @@ extension FBSnapshotTestCase {
         activationPointDisplayMode: ActivationPointDisplayMode,
         useMonochromeSnapshot: Bool
     ) -> String? {
+        // The tests must be run in a host application in order for the accessibility properties to be populated
+        // correctly. The `UIApplication.shared` singleton is non-optional, but will be uninitialized when the tests are
+        // running outside of a host application, so we can use this check to determine whether we have a test host.
+        let hostApplication: UIApplication? = UIApplication.shared
+        guard hostApplication != nil else {
+            return "Accessibility snapshot tests cannot be run in a test target without a host application"
```
dfed, line 57: Asking to learn: why are we returning an error message here? Seems like a reasonable error message but I’m having trouble understanding why it’s being returned as a string.

AccessibilitySnapshot/Classes/FBSnapshotTestCase+ObjCSupport.swift
```diff
@@ -49,6 +49,14 @@ extension FBSnapshotTestCase {
         activationPointDisplayMode: ActivationPointDisplayMode,
         useMonochromeSnapshot: Bool
     ) -> String? {
+        // The tests must be run in a host application in order for the accessibility properties to be populated
+        // correctly. The `UIApplication.shared` singleton is non-optional, but will be uninitialized when the tests are
+        // running outside of a host application, so we can use this check to determine whether we have a test host.
+        let hostApplication: UIApplication? = UIApplication.shared
+        guard hostApplication != nil else {
```
dfed, line 56: Since we’re making this check more than once: thoughts on pulling out this check and associated comments into a method we can call in both places?

AccessibilitySnapshot/Classes/FBSnapshotTestCase+ObjCSupport.swift
NickEntin: This is how the Objective-C test assertions put the error on the correct line. The actual test code is a method that returns a `String?`. The Objective-C tests use a macro that calls this method. If the method returns nil, it considers the test to have succeeded. If the method returns a string, it considers the test to have failed and calls `XCTFail()` with that as the failure message. Since it's a macro, the failure shows up on that line, rather than in the framework code.

dfed: Got it! Thanks for the explanation.

AccessibilitySnapshot/Classes/FBSnapshotTestCase+ObjCSupport.swift
NickEntin: Moved into a shared `isRunningInHostApplication` property 👍

AccessibilitySnapshot/Classes/FBSnapshotTestCase+Accessibility.swift
NickEntin: I could go either way on this, but I think it's still helpful to see which snapshot test failed. For example, if you want to temporarily disable the verify calls that are failing, it's helpful to highlight which ones are requiring the test host (rather than at a test level, since you could mix standard snapshot calls and accessibility snapshot calls in the same test). Given the descriptive failure message, I don't think it's that useful to know where in the framework the problem occurred, since ideally you don't have to know anything under the public API.

Going to merge as-is for now, but happy to keeping chatting and follow-up with a change if you disagree.

dfed: Got it. I was imagining a world where accessibility snapshot tests were running as their own target. Makes sense that they might be mixed.

Issue #15, opened 2020-08-04
Improve support for snapshotting views with text field cursors to prevent tests from flaking


allisonmoyer:

NickEntin:
Updated the title to be less prescriptive about the solution to this problem. The root problem here is that snapshotting views that have layer animations can lead to flaky tests, since the view may be snapshotted at different points in the animation across different test runs. I can see a few solutions to this:

* **Hide the text field cursor.** This can be done by modifying the tint color (what's in #16), using private APIs to set the cursor view to hidden, etc.
* **Pause the layer animation.** We can set the layer's `speed` to 0 and the `timeOffset` to some fixed number to pause the animation at a specific point. Depending on what `timeOffset` we use, we can set the cursor to be hidden or visible.

The second approach could potentially apply much broader than text fields, since this same problem exists for any layer that has an animation applied to it. We might want a targeted solution for now though. In either case, I think we should make sure we:

* Make it obvious what the behavior will be and make it easy to opt out. Consumers may be using the `tintColor` other places in their code, or already set layer animations to a specific point, so we should let them control this behavior.
* Restore the original state after snapshotting the view. Since there may be multiple snapshot calls (or other assertions) in each test method, we should make sure to restore any state that we set before the snapshot method returns.

fbernutz:
Another idea might be to use the `precision` value which is used in `swift-snapshot-testing` to prevent flaky tests for things like cursors in textfields. For some reason in the project I'm currently working in (where we use this a11y snapshot library), we need a 99,9% of precision to make the tests pass in the CI, so to have the ability to adjust the precision would be a great advantage.

NickEntin:
I think it's still worth solving the cursor problem separately from adding support for specifying a `precision` for the comparison.

[...]

PR #16, opened 2020-08-04
Hide the cursor on text fields to fix flaky behavior of tests

allisonmoyer:
Fixes issue #15


Example/SnapshotTests/TextFieldCursorTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+        let container = ContainerView(subview: textField)
+        container.frame.size = container.sizeThatFits(UIScreen.main.bounds.size)
+
+        SnapshotVerifyAccessibility(container)
+
+        // Since this is fixing flaky behavior, call to `SnapshotVerifyAccessibility`
+        // may not fail if this regresses, so also explicitly test that the call to
+        // `SnapshotVerifyAccessibility` cleared the tint color of the text field.
+        XCTAssertEqual(textField.tintColor, .clear)
```
NickEntin, line 36: Calling `SnapshotVerifyAccessibility(...)` should not mutate the view being snapshotted. There may be multiple snapshot calls in a single test, or other assertions, and having side effects in one may break the others.

It's okay to temporarily change values inside the snapshot method, but we should restore the original values before returning.

PR #18, opened 2020-08-23
Include custom actions in accessibility snapshots

### LINKED ISSUE #10: Add custom actions to accessibility markers
There are two main chunks of work for supporting custom actions:

1. Add custom actions to the `AccessibilityMarker` struct and populate them when parsing the accessibility hierarchy.

2. Show a list of custom actions in the `AccessibilitySnapshotView.LegendView`.

NickEntin:
Resolves #10.


AccessibilitySnapshot/Core/Classes/AccessibilitySnapshotView.swift
```diff
@@ -412,17 +412,36 @@ private extension AccessibilitySnapshotView {
                 return label
             }
 
+            // If our description and hint are both empty, but we have custom actions, we'll use the description label
+            // to show the "Actions Available" text, since this makes our layout simpler when we align to the marker.
+            let showActionsAvailableInDecription = (marker.description.isEmpty && !marker.customActions.isEmpty)
```
lukebradford, line 417: `Decription` -> `Description`

AccessibilitySnapshot/Core/Classes/AccessibilitySnapshotView.swift
```diff
@@ -518,12 +553,162 @@ private extension AccessibilitySnapshotView {
...
+    // MARK: -
+
+    private final class CustomActionsView: UIView {
+
+        // MARK: - Life Cycle
+
+        init(actionsAvailableText: String?, customActions: [String]) {
+            actionLabels = customActions.map {
+                let iconLabel = UILabel()
+                iconLabel.text = "↓"
```
lukebradford, line 586: Why use a down arrow for this?

NickEntin: I'm trying to reflect how you interact with the element. It will immediately say "Actions Available." You can then perform an up or down flick gesture to iterate through the actions. They're ordered here the same as flicking down through the list (which also matches the order of the actions in the `accessibilityCustomActions` array.

AccessibilitySnapshot/Core/Classes/AccessibilitySnapshotView.swift
NickEntin: Good catch!

NickEntin:
Screen Shot 2020-08-26 at 3 25 22 PM
😑  Those are referring to the same build. It passed, but for some reason it's not informing GitHub of the updates correctly.

PR #23, opened 2020-09-25
Point demo app at master branch of Paralayout

NickEntin:
This updates the demo app's dependency on Paralayout to point at the `master` branch to fix a build issue in Xcode 12. This is not a dependency in AccessibilitySnapshot itself, so it should be safe to not be using a release version.


seanho (approved):
LGTM. Would be nice to link back to the original issue in Paralayout.

Issue #1, opened 2020-05-19
Add support for installing via Swift Package Manager


NickEntin:

NickEntin:
In order to add SPM support for AccessibilitySnapshot, we need its dependencies to work with SPM first. There's an open issue for iOSSnapshotTestCase to add support (uber/ios-snapshot-test-case#91) and adding it for fishhook should be relatively straightforward.

Sherlouk:
Any progress on this Nick? I understand there's a minor dependency on Uber (though a MR is open) but I'm unsure how you wanted to proceed with fishhook -- would be nice to get Core compatible with SPM at least!

NickEntin:
Hey James, no progress on this yet. As you said, fishhook is the only dependency needed to add SPM support to the core framework. I filed facebook/fishhook#77 to make sure it's being tracked for that project.

Sherlouk:
🙌 Merge Request is open at https://github.com/facebook/fishhook/pull/78

We've got a few projects which are SPM only so looking forward to being able to support this for SnapshotTesting

Sherlouk:
@NickEntin 

I'm not sure if the fishhook repo is being actively monitored/managed etc by the teams at Facebook - how against the idea of using the fork which has support for it are you?

Would be awesome to unblock the SPM support here especially as teams (including mine!) are transitioning to SPM!

NickEntin:
I commented on the fishhook PR and tagged a couple of reviewers from the most recent merged PRs to see if they can take a look. I'd love to see SPM support here unblocked soon. 🙌 

My main concern with switching over to a fork is that (presumably) it would be published as a separate pod, which could potentially be problematic if consumers of AccessibilitySnapshot already use fishhook for something else (either directly or as a transitive dependency). I think @dfed mentioned a while back that there are some workarounds we could do in SPM to deal with a missing dependency, but let's give fishhook another push first and see if we can get it supported properly.

Sherlouk:
Hey @NickEntin 

If you get 15 minutes I don't support you'd mind taking a gander at my very WIP branch for Swift Package Manager support?

https://github.com/Sherlouk/AccessibilitySnapshot/tree/sherlouk/spm

The branch is unorganised so don't worry too much about how the pods stuff is setup (Should be no problem keeping them both working at the same time, even without changing the dependencies for pod consumers).

[...]

PR #24, opened 2020-09-27
Add Swift Package Manager Support

Sherlouk:
⚠️ Work in progress...

Based on conversation in #1 
This is also extending from the work in #21 enabling SnapshotTesting support (which is the default for Swift Package Manager installation).

Changes:
- Restructured repository so source code more easily followed the Swift Package format. This includes pulling public headers out into an `include` directory which simplifies setup.
- Updated Podspec to use the new structure ensuring.
- Moved UnitTests to the package, rather than in the example project.
- Duplicated the entire SnapshotTests scheme, replacing the Uber syntax with Pointfree's syntax. I was tempted to try and have the SnapshotTesting subspec use the same suite of tests to ensure both installation methods have the exact same output, but didn't run with this yet.
- Increased deployment target to iOS 11

Notes:
- The example project doesn't like having both the CocoaPods version and the Swift Package Manager version building at the same time. As it stands, you need to enabling one or the other via the scheme but both do work when run individually.
- I have not included the snapshots for the Swift Package Manager target, though I have verified everything seems to be working as expected.
- This is currently using a forked version of fishhook for Swift Package Manager only. The CocoaPods installation is completely unaffected. If we continue with this then it'd make sense to open a PR for my version of fishhook into the parent repository -- but we could easily and seamlessly change this dependency at a later date without really any bother to our users.

TODO:
- Find workaround for the CocoaPods+SPM issue noted above.
- Figure out how to implement the `SnapshotTestCase` for SnapshotTesting. SnapshotTesting doesn't (yet, there have been many discussions...) support adding simulator details such as OS, screen size or screen scale to the image name like Uber's library does.
- CI Fixed
- Documentation Changes

@NickEntin - would love to hear your opinions on the work here and particularly if you have any thoughts on how to approach the two top TODOs and honestly listing all the issues with the work here that you'd want changing! Thanks 🙌


NickEntin:
This is looking great so far!

> The example project doesn't like having both the CocoaPods version and the Swift Package Manager version building at the same time. As it stands, you need to enabling one or the other via the scheme but both do work when run individually.

That's fine. We can have two schemes - one for each of the dependency managers, something like `AccessibilitySnapshotDemo (FBSnapshotTestCase)` and `AccessibilitySnapshotDemo (SnapshotTesting)`. We don't need to test both using the same scheme.

> Figure out how to implement the SnapshotTestCase for SnapshotTesting. SnapshotTesting doesn't (yet, there have been many discussions...) support adding simulator details such as OS, screen size or screen scale to the image name like Uber's library does.

Hmm, I don't have enough experience with SnapshotTesting to know what all options are available here. @efirestone do you have any ideas how we can do this?

Sherlouk:
Okay so both schemes are happily working locally on Xcode 12 but CI is having a cry because.. well.. it's not on Xcode 12. Also got CocoaPods lint all happy 👍 

We might need to make an entirely separate project to get the Swift Package version working on CI. Thoughts?

Will wait on guidance around how we want to handle the different scales as well.

NickEntin:
Should be able to rebase on `master` now to avoid the `fishhook` dependency

Sherlouk:
Awesome! I'll take a fresh branch and go at it again this week (hopefully 🤞) - I'll start a new draft PR and I'll leave out most of the changes to tests (for now) focussing on the restructure and basic SPM support 👍

PR #28, opened 2020-10-07
Switch default subspec to SnapshotTesting

### LINKED ISSUE #22: Switch default snapshot engine over to SnapshotTesting
We should change the default subspec from `FBSnapshotTestCase` to `SnapshotTesting`. `SnapshotTesting` has a deployment target of iOS 11, while we're currently at iOS 10, so it probably makes sense to bump our deployment target in parallel to this change.

NickEntin:
Resolves #22.

cc @Sherlouk

### DIFF
--- a/Example/Podfile
+++ b/Example/Podfile
@@ -10,7 +10,7 @@ target 'AccessibilitySnapshotDemo' do
 	target 'SnapshotTests' do
 		inherit! :search_paths
 
-		pod 'AccessibilitySnapshot', :path => '../AccessibilitySnapshot.podspec'
+		pod 'AccessibilitySnapshot/iOSSnapshotTestCase', :path => '../AccessibilitySnapshot.podspec'
 		pod 'AccessibilitySnapshot/SnapshotTesting', :path => '../AccessibilitySnapshot.podspec'
 	end
 
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,7 +1,4 @@
 PODS:
-  - AccessibilitySnapshot (0.3.3):
-    - AccessibilitySnapshot/Core (= 0.3.3)
-    - AccessibilitySnapshot/iOSSnapshotTestCase (= 0.3.3)
   - AccessibilitySnapshot/Core (0.3.3):
     - fishhook (~> 0.2)
   - AccessibilitySnapshot/iOSSnapshotTestCase (0.3.3):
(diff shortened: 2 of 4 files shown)


Example/Podfile.lock
```diff
@@ -17,11 +14,11 @@ PODS:
   - iOSSnapshotTestCase/SwiftSupport (6.2.0):
     - iOSSnapshotTestCase/Core
   - Paralayout (0.9.1)
-  - SnapshotTesting (1.7.2)
+  - SnapshotTesting (1.8.2)
```
NickEntin, line 17: Looks like this version of SnapshotTesting doesn't build with Xcode 10. I don't think we're getting anything from Xcode 10 support at this point, so probably fine to update CI to Xcode 11-only.

Example/Podfile
```diff
@@ -10,7 +10,7 @@ target 'AccessibilitySnapshotDemo' do
 	target 'SnapshotTests' do
 		inherit! :search_paths
 
-		pod 'AccessibilitySnapshot', :path => '../AccessibilitySnapshot.podspec'
+		pod 'AccessibilitySnapshot/iOSSnapshotTestCase', :path => '../AccessibilitySnapshot.podspec'
 		pod 'AccessibilitySnapshot/SnapshotTesting', :path => '../AccessibilitySnapshot.podspec'
```
Sherlouk, line 14: Technically we don't need the `/SnapshotTesting` part here but equally might be worth keeping to be explicit about which specs we are using?

README.md
```diff
@@ -9,7 +9,7 @@ AccessibilitySnapshots makes it simple to add regression tests for accessibility
 
 ## Getting Started
 
-By default, AccessibilitySnapshot uses [iOSSnapshotTestCase](https://github.com/uber/ios-snapshot-test-case) to record snapshots and perform comparisons. Before setting up accessibility snapshot tests, make sure your project is set up for standard snapshot testing. Accessibility snapshot tests require that the test target has a host application. See the [Extensions](#extensions) section below for a list of other available snapshotting options.
+By default, AccessibilitySnapshot uses [SnapshotTesting](https://github.com/pointfreeco/swift-snapshot-testing) to record snapshots and perform comparisons. The framework also includes support for using [iOSSnapshotTestCase](https://github.com/uber/ios-snapshot-test-case) as the snapshotting engine instead. Before setting up accessibility snapshot tests, make sure your project is set up for standard snapshot testing. Accessibility snapshot tests require that the test target has a host application. See the [Extensions](#extensions) section below for a list of other available snapshotting options.
```
Sherlouk, line 12: Do we think it's worth adding a note in the README under this paragraph something across the lines of:

> ⚠️  Updating to v0.4.0
>
> Until v0.4.0 AccessibilitySnapshot defaulted to using [iOSSnapshotTestCase](/). If you are upgrading from an older version of AccessibilitySnapshot then you will need to update your Podfile to use `pod 'AccessibilitySnapshot/iOSSnapshotTestCase'` instead of `pod 'AccessibilitySnapshot'`.

This equally might work in an MIGRATION.md or something which we could update in the future for breaking changes?

Sherlouk (approved):
Looks good 👍

README.md
NickEntin: I put it in the release notes for 0.4.0. Hopefully that would be one of the first places people would look to see what's new in the release that broke for them.

Screen Shot 2020-10-16 at 8 55 21 PM

Example/Podfile
NickEntin: Yeah, I think it's probably good to be explicit about it here. One less change to make if/when we add a new subspec or change the default subspecs.

NickEntin:
This is currently blocked on #40

Example/Podfile.lock
NickEntin: Bumped our CI to Xcode 11 across the board. Should be good now.

efirestone (approved):

PR #30, opened 2020-10-07
Revert bump to iOS 12 deployment target

### LINKED ISSUE #29: CI builds for iOS 13 are flaking
The CI build on iOS 13 seems to be flaking on a few specific tests:

```diff
testTabBars()
- Item B: 3 items. Tab. 2 of 4.
+ Item B. Tab. 2 of 4.
- Item C: A. Tab. 3 of 4.
+ Item C. Tab. 3 of 4.

testStepper()
- Decrement. Button.
+ minus. Button.
- Increment. Button.
+ plus. Button.

testStepperAtMin()
- Decrement. Dimmed. Button.
+ minus. Dimmed. Button.
- Increment. Button.
+ plus. Button.
```

This started after bumping the minimum iOS version to 12.0 (#27). I've seen this flake twice now.

NickEntin:
Testing this out to see if it fixes #29, since the flaky builds were first seen after this change.

Running tally:
| Passed | Failed |
| :---: | :---: |
| 4 | 0 |


NickEntin:
4 successful builds in a row is making me think this is definitely what broke it. Still no idea why though. 😕

NickEntin:
I'm guessing this is from the deployment target change, but it's possible that I subtly changed something accidentally when removing the availability checks. Testing that now...

Running tally:
| Passed | Failed |
| :---: | :---: |
| 0 | 0 |

NickEntin:
Looks like the same tests are flaking

```
Failing tests:
	SnapshotTests:
		AccessibilitySnapshotTests.testTabBars()
		DefaultControlsTests.testStepper()
		DefaultControlsTests.testStepperAtMin()
```

NickEntin:
Testing with the framework's minimum version set to iOS 10, but the demo app's deployment target (which is shared with the tests) to iOS 12.

Running tally:
| Passed | Failed |
| :---: | :---: |
| 0 | 0 |

NickEntin:
The build seems pretty stable. This implies that the framework's minimum supported version is what matters, not the app's deployment target.

I'm not sure why that version change makes a difference, but unfortunately it means that bumping the version will decrease test stability for consumers.

PR #31, opened 2020-10-09
Change minimum supported iOS version to 11.0

### LINKED ISSUE #29: CI builds for iOS 13 are flaking

NickEntin:
Testing this out to see if it fixes #29. If this works, it means that having the minimum version at 12.0 is the issue, and using 11.0 would be an option until we figure out the root cause of the flaky tests.

Running tally:
| Passed | Failed |
| :---: | :---: |
| 4 | 0 |

### DIFF
--- a/AccessibilitySnapshot.podspec
+++ b/AccessibilitySnapshot.podspec
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
 
   s.swift_version = '5.0.1'
 
-  s.ios.deployment_target = '12.0'
+  s.ios.deployment_target = '11.0'
 
   s.default_subspecs = 'Core', 'iOSSnapshotTestCase'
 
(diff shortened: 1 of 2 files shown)


NickEntin:
Build seems pretty stable. This implies that iOS 12 as the base SDK is the issue.

PR #32, opened 2020-10-12
Create Accessibility Description SnapshotTesting Strategy

Sherlouk:
Sometimes a user may wish not to snapshot the visual part of their app as part of their accessibility tests. This strategy would give those users an alternative option using a textual representation of just the VoiceOver capability.


NickEntin:
@Sherlouk I'm gonna close this PR for now as we discuss text-based strategies in #93. We can reopen with the appropriate changes once we figure out the full plan.

Issue #33, opened 2020-10-15
SwiftUI Views with ScrollViews for iOS 14 don't record Voice Over Descriptions


fbernutz:
Something seems to have changed from iOS 13 to iOS 14 with SwiftUI ScrollViews. 

When snapshotting a SwiftUI View with a ScrollView (only iOS 14), no Accessibility Nodes can be found. I tried to debug this and found differences in the objects when the SwiftUI View had a ScrollView and when it had no ScrollView.

iOS 13: Elements inside a ScrollView are of type NSObject
<details>
    <summary>Screenshot</summary>

   <img width="1792" alt="Screen Shot 2020-10-14 at 09 33 54" src="https://user-images.githubusercontent.com/26111180/96107537-d095bc00-0edc-11eb-92a8-d97bd5ec537c.png">

</details>

iOS 14: Elements inside a ScrollView are of type UIView (SwiftUI_UIShapeHitTestingView)
<details>
    <summary>Screenshot</summary>

<img width="1792" alt="Screen Shot 2020-10-14 at 09 31 24" src="https://user-images.githubusercontent.com/26111180/96107559-d68b9d00-0edc-11eb-8502-3b957ea1c793.png">

</details>

This results in no accessibility nodes for the iOS 14 snapshot. 

| iOS 13 | iOS 14 |
|---|---|
| ![testSimpleSwiftUIWithScrollViewConfiguration 375x812-13-3-3x](https://user-images.githubusercontent.com/26111180/96107663-f91db600-0edc-11eb-8fef-f340f644af06.png) | ![testSimpleSwiftUIWithScrollViewConfiguration 375x812-14-0-3x](https://user-images.githubusercontent.com/26111180/96107676-fb801000-0edc-11eb-936c-e5dfad228f1f.png) |

I've tried to fix this but honestly I have no idea how 🙈 

I've also added a draft PR #34 with two new snapshot tests where you can see the resulting snapshot for these scenarios. If there are any more information needed to look into this, I'm happy to help.

NickEntin:
Thanks for the report @fbernutz! I'll run through it in the debugger and see if I can figure out what's going on.

NickEntin:
[...]

PR #34, opened 2020-10-15
Add SwiftUI View to SnapshotTests

fbernutz:
This adds two SwiftUI Views as example for issue #33


Example/AccessibilitySnapshot/SwiftUIView+EmbedInHostingController.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//  Created by Felizia Bernutz on 15.10.20.
+//  Copyright © 2020 CocoaPods. All rights reserved.
+//
+
+import UIKit
+import SwiftUI
+
+@available(iOS 13.0, *)
+extension View {
```
Sherlouk, line 13: This is definitely an acceptable solution - but it may also we be worth us adding SwiftUI support to the SnapshotTesting spec at some point!

The base library does support it so would remove this extension being necessary for us or users of the library!

fbernutz: yes definitely, would be great 👍

fbernutz: I've added a comment to remember to remove this extension again when #37 is fixed and snapshotting SwiftUI views is supported. Hope that's fine for the moment.

fbernutz:
As Issue #33 seems to be a SwiftUI bug in iOS 14.0 and 14.1, this branch might not be needed anymore. But if you want, let's use it as a preparation for #37 because there are now two SwiftUI views which can be used for snapshot tests. And we now know when the SwiftUI issue is present again in newer iOS versions because we have a specific test for that.

Example/AccessibilitySnapshot/SwiftUIView+EmbedInHostingController.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  Copyright 2019 Square Inc.
```
NickEntin, line 2: 2021 🙃 

```suggestion
//  Copyright 2021 Square Inc.
```

Example/AccessibilitySnapshot/SwiftUIView+EmbedInHostingController.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+import SwiftUI
```
NickEntin, line 18: style nit: sort the imports

```suggestion
import SwiftUI
import UIKit
```

Example/AccessibilitySnapshot/SwiftUIViewWithScrollView.swift
```diff
@@ -0,0 +1,38 @@
...
+//  limitations under the License.
+//
+
+import SwiftUI
+
+@available(iOS 13.0, *)
+/// A SwiftUI View inside a ScrollView will produce no accessibility elements
+/// for iOS 14.0 and 14.1, for more information see the issue:
+/// AccessibilitySnapshot/issues/33. This seems to be a SwiftUI bug which was
+/// resolved in iOS 14.2.
```
NickEntin, line 23: nit: Let's format as `cashapp/AccessibilitySnapshot#33` since that's what GitHub will automatically link if you copy it. Also, I think the comment can be above the `@available` to keep the code together.

```suggestion
/// A SwiftUI View inside a ScrollView will produce no accessibility elements for iOS 14.0 and 14.1, for more
/// information see cashapp/AccessibilitySnapshot#33. This seems to be a SwiftUI bug which was resolved in iOS 14.2.
@available(iOS 13.0, *)
```

Example/AccessibilitySnapshot/SwiftUIView+EmbedInHostingController.swift
```diff
@@ -0,0 +1,44 @@
...
+
+@available(iOS 13.0, *)
+extension View {
+
+    // should be removed again when AccessibilitySnapshot supports snapshotting
+    // native SwiftUI views, see AccessibilitySnapshot/issues/37
```
NickEntin, line 24: nit: formatting

```suggestion
    // This method should be removed once we support snapshotting native SwiftUI views.
    // See cashapp/AccessibilitySnapshot#37.
```

Example/AccessibilitySnapshot/SwiftUIView+EmbedInHostingController.swift
```diff
@@ -0,0 +1,44 @@
...
+    func toVC() -> UIViewController {
+        let viewController = UIViewController()
```
NickEntin, line 26: Why do we need to wrap the hosting controller in another VC?

Example/AccessibilitySnapshot/SwiftUIViewWithScrollView.swift
fbernutz: good to know 👌

Example/AccessibilitySnapshot/SwiftUIView+EmbedInHostingController.swift
fbernutz: mh 🤔  good point, will try it out if it's possible to snapshot the HostingController directly

fbernutz: I removed the Container and it still results in the same Snapshot! 👍  Great finding!

NickEntin (approved):
Looks great! Thanks for adding this 🙌

PR #36, opened 2020-10-16
Debug flaky parser tests

NickEntin:
Adds a bunch of logs to help debug #29

### DIFF
--- a/AccessibilitySnapshot/Core/Classes/AccessibilityHierarchyParser.swift
+++ b/AccessibilitySnapshot/Core/Classes/AccessibilityHierarchyParser.swift
@@ -150,6 +150,8 @@ public final class AccessibilityHierarchyParser {
 
     // MARK: - Public Methods
 
+    public var mostRecentNodeDescriptions: [String] = []
+
     /// Parses the accessibility hierarchy starting from the `root` view and returns markers for each element in the
     /// hierarchy, in the order VoiceOver will iterate through them when using flick navigation.
     ///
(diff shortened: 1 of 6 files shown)


NickEntin:
Hmm, 4 builds in a row have passed now. 🤔 Something in the debugging code resolved the issue.

NickEntin:
So it definitely seems like the log statements in the parser are what's fixing the build. Which... definitely shouldn't fix the build. Just to make sure there's nothing else going on...

NickEntin:
So even something as simple as adding a test attachment is causing the build to be stable. I'm thinking this must be a very slight timing issue, although I can't think of why the timing would matter here.

NickEntin:
Testing out replacing the print statement with a call to `fflush`, per @efirestone 's suggestion.


Running tally:
| Passed | Failed |
| :---: | :---: |
| 1 | 0 |

NickEntin:
Testing with a totally empty `print` statement


Running tally:
| Passed | Failed |
| :---: | :---: |
| 1 | 0 |

NickEntin:
Trying something that should definitely be a no-op


Running tally:
| Passed | Failed |
| :---: | :---: |
| 1 | 0 |

NickEntin:
Original code, different SHA

Running tally:
| Passed | Failed |
| :---: | :---: |
| 1 | 0 |

NickEntin:
Checking whether changing something in the test target will fix this, meaning that the problem is in the test target rather than the framework itself

Running tally:
| Passed | Failed |
| :---: | :---: |
| 1 | 0 |

NickEntin:
Since a change in the test target seems to make the build stable, it feels like the problem is actually in the test target and not the framework itself.

PR #38, opened 2020-10-26
Split CI jobs into separate workflows

NickEntin:
Since GitHub Actions doesn't allow for restarting a single job, this breaks our large CI workflow into separate workflows for each of the job types. This makes it easier to debug flaky builds without having to restart as many jobs.


NickEntin:
This is helpful, but it might have some odd implications (like how we report the status on the README). Declining for now.

PR #39, opened 2020-10-29
Bump SnapshotTesting dependency in sample app to 1.8.2

NickEntin:
Now that we have a deployment target of iOS 12, we can test against the latest version of SnapshotTesting.

cc @Sherlouk


NickEntin:
Ahh right, SnapshotTesting 1.8.2 doesn't support Xcode 10, so it's failing on Travis. 🤦 As a side note, that feels like it should have been a breaking change in SnapshotTesting. In any case, I think it's alright for us to drop Xcode 10 support in 0.4, but there's some other CI work we need to do first to unblock that.

NickEntin:
This was already handled as part of #28

PR #42, opened 2020-10-30
Replace dependency on fishhook with inlined code

NickEntin:
Fishhook is currently blocking SPM support (#1) and hasn't been merging PRs recently, so it looks like that support likely isn't coming soon. This removes our dependency on Fishhook by inlining the code that we use from it. If Fishhook ends up getting more updates in the future, we can consider re-adding it as a dependency.

Probably easiest to review by commit. The first commit copies in the code from Fishhook and the remaining commits extract out parts of it that we don't need.


NickEntin:
cc @Sherlouk

Sherlouk:
Awesome stuff!

efirestone (approved):

PR #46, opened 2020-11-11
Fix edge cases around nested accessibility containers

NickEntin:


NickEntin:
Turns out the behavior here was already correct. We may want to add these tests as a form of documentation and/or regression testing, but they probably need a bit more iteration from what's here. Closing for now.

PR #47, opened 2020-11-13
Add Swift Package Manager

Sherlouk:
Supersedes #24 

* Removed `AccessibilitySnapshot` directory in favour of a `Sources` directory (which is closer to the Swift Package Manager expected format)
* Updated the CocoaPods installation method to point to the new file locations
* Minor tweaks to source files to make them work with Swift Package Manager (uses compiler flags as to allow CocoaPods support to remain unaffected)
* Added Package.swift!! 🙌 

Incomplete:
- [x] Update Documentation

---

Tried to keep the changes as small as possible in this first merge request allowing us to iterate more easily!

Project builds correctly when opening the Swift Package (but no tests). Have also run `pod install` and then run the tests in the Example project and they're working fine for me. (Let's see how CI does!)

@NickEntin - Let me know what you think on these changes 😄


.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
NickEntin, line 1: Let's add the standard SPM stuff to the `.gitignore`
```
# Swift Pacakge Manager
generated/
.build/
.swiftpm/
```

AccessibilitySnapshot.podspec
```diff
@@ -15,33 +15,30 @@ Pod::Spec.new do |s|
   s.default_subspecs = 'Core', 'SnapshotTesting'
 
   s.subspec 'Core' do |ss|
-    ss.source_files = 'AccessibilitySnapshot/Core/Classes/**/*.{swift,h,m}'
-    ss.public_header_files = [
-      'AccessibilitySnapshot/Core/Classes/UIAccessibilityStatusUtility.h',
-      'AccessibilitySnapshot/Core/Classes/UIView+DynamicTypeSnapshotting.h',
-    ]
+    ss.source_files = 'Sources/AccessibilitySnapshotCore/Classes/**/*.swift', 'Sources/AccessibilitySnapshotCore-ObjC/**/*.{h,m}'
+    ss.public_header_files = 'Sources/AccessibilitySnapshotCore-ObjC/include/*.h'
```
NickEntin, line 19: Can all of the source files be in `Sources/AccessibilitySnapshot/Core/` (or `Sources/AccessibilitySnapshotCore`) and differentiate in SPM based on the file type? Feels really weird to separate these out when they're conceptually the same framework.

Sherlouk: SPM seems to struggle with mixed language targets/source file folders. This was one of the nicer ways I was able to get it all working!

It should be possible to do:
`Sources/AccessibilitySnapshot/Core/ObjC[/include]`
`Sources/AccessibilitySnapshot/Core/Swift`
If you think that's better?

NickEntin: Hmm, that might work. Totally understand the limitation around mixed targets. I was suggesting breaking the strict target/directory pairing and explicitly defining the source file pattern, although it looks like SPM only supports an explicit source list (no glob matching). 😞

I think splitting in `/ObjC` and `/Swift` would be cleaner here though 👍

Sherlouk: Yeah the lack of glob matching is a bit sad 😭  I'll work on splitting it up like that!

Sherlouk: Should be resolved 👍  Sorry for the delay

Sherlouk:
Awesome looks like CI is happy. Care to take a look when you're not busy @NickEntin? 🙏  Thanks!

I can also update some docs once you're happy

.gitignore
```diff
@@ -27,3 +27,8 @@ Carthage/Build
 
 # CocoaPods
 Pods/
+
+# Swift Package Manager
+generated/
+.build/
+.swiftpm/
```
NickEntin, line 34: nit: Let's leave the newline at the end
```suggestion
.swiftpm/

```

Sources/AccessibilitySnapshot/SnapshotTesting/SnapshotTesting+Accessibility.swift
```diff
@@ -14,8 +14,14 @@
 //  limitations under the License.
 //
 
+import UIKit
 import SnapshotTesting
```
NickEntin, line 18: nit: Sort the imports
```suggestion
import SnapshotTesting
import UIKit
```

Package.swift
```diff
@@ -0,0 +1,47 @@
+// swift-tools-version:5.3
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "AccessibilitySnapshot",
+    defaultLocalization: "en",
+    platforms: [
+        .iOS(.v11),
```
NickEntin, line 10: Let's make it 12 to match the current CocoaPods config. I think it will be easier to have the same minimum version across the board.
```suggestion
        .iOS(.v12),
```

NickEntin (commented):
This is looking great! Just a few small comments plus adding install instructions to the README, and I think this should be good to go

Package.swift
```diff
@@ -0,0 +1,47 @@
+// swift-tools-version:5.3
```
NickEntin, line 1: Do we actually need 5.3 for this? I think we could probably go back to 5.0.1 and be fine.

NickEntin:
Oh, we should also add a CI build for this. I pushed a [commit](https://github.com/NickEntin/AccessibilitySnapshot/commits/entin/spm-ci) to add the CI build, but I'm getting errors when I run it locally.

Package.swift
Sherlouk: The syntax required to install SnapshotTesting doesn't exist that far back. Earliest I could find was 5.2.
Also we require the `defaultLocalization` due to the strings files the project contains - this wasn't added until 5.3 (see [here](https://github.com/apple/swift-evolution/blob/master/proposals/0278-package-manager-localized-resources.md))

Sherlouk:
Done some basic docs - just need to get the CI build working 👍

Sherlouk:
Pushed up one change which seems to make it a lot happier than it was... 

I'm also trying to run the build via the Swift command itself running:

```
swift build -Xswiftc "-sdk" -Xswiftc "`xcrun --sdk iphonesimulator --show-sdk-path`" -Xswiftc "-target" -Xswiftc "x86_64-apple-ios14.0-simulator" 
```

but it seems to be freaking out on the lack of UIKit. All very peculiar because in a normal project it runs absolutely fine!

One idea might be to setup an actual Xcode project which uses the library via a local dependency? If we can build/test that (add one basic example?) then it would prove what we need it to and circumvent the direct building stupidities?

Package.swift
NickEntin: Cool, leaving it at 5.3 sounds good then 👍🏻

Sherlouk:
Spoke to someone at work who knows way more about Swift than me and he pointed me down the right track!

tl;dr `swift package generate-xcodeproj` should be avoided. If you don't specify a project then xcodebuild will pickup the Package.swift and do what it's needs to do.

All working locally using iOS 14

NickEntin:
Ahh, interesting! I didn't know that `xcodebuild` would automatically pick up a `Package.swift`. It still seems odd to me that generating an Xcode project doesn't work though. It looks like there's two general sets of errors. The first is that `Bundle.module` can't be found, which is fixed by specifying the resources in the package definition:

```diff
         .target(
             name: "AccessibilitySnapshotCore",
             dependencies: ["AccessibilitySnapshotCore-ObjC"],
-            path: "Sources/AccessibilitySnapshot/Core/Swift"
+            path: "Sources/AccessibilitySnapshot/Core/Swift",
+            resources: [.process("Assets")]
         ),
```

The second set of errors is around undefined symbols in UIKit. That one I'm not sure about. Any idea why it wouldn't linking UIKit correctly when generating the Xcode project? Avoiding generating the Xcode project in CI seems reasonable, but we should make sure this isn't revealing some other underlying issue.

Sherlouk:
@NickEntin from my understanding this is just a bug with the project that gets generated - a lot of the flags that are needed for an iOS target haven't been configured correctly (bare in mind that iOS support is relatively new to SPM!)

I have tried this exhaustively in projects and haven't seen any projects - I'm pretty confident what we were seeing on CI was just an issue with the approach (and not something you'd do in a standard iOS project)

Sherlouk:
I had (and can push it to the project if you'd like) made a completely separate CI iOS Xcode project which just showed that including the package normally and using it in a test worked absolutely fine.

This might be another option if you're not confident with the current CI approach!

.github/workflows/ci.yml
```diff
@@ -43,3 +43,17 @@ jobs:
...
+      matrix:
+        platform: ['iOS_14']
+      fail-fast: false
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Select Xcode Version
+        run: sudo xcode-select --switch /Applications/Xcode_12.2.app/Contents/Developer
+      - name: Build
+        run: Scripts/build.swift spm ${{ matrix.platform }}
```
NickEntin, line 59: nit: Leave a newline at the end of the file
```suggestion
        run: Scripts/build.swift spm ${{ matrix.platform }}

```

AccessibilitySnapshot.podspec
```diff
@@ -15,33 +15,30 @@ Pod::Spec.new do |s|
...
+    ss.source_files = 'Sources/AccessibilitySnapshot/SnapshotTesting/**/*.{swift,h,m}'
 
     ss.dependency 'AccessibilitySnapshot/Core'
     ss.dependency 'SnapshotTesting', '~> 1.0'
   end
 
   s.frameworks = 'XCTest'
   s.weak_frameworks = 'XCTest'
-end
+end
```
NickEntin, line 44: nit: Leave the newline at the end of the file
```suggestion
end

```

NickEntin (approved):

NickEntin:
Got it. As long as the test project was working for you, I think this should be fine. 👍 

Just a couple small nits, and then I think we're good to merge!

Sherlouk:
As I say the purpose of CI is to give us confidence -- more than happy to explore other changes and improvements. Or we can just wait for the first issue to be raised and discuss how wrong I am in there 😅 

Nits all merged 🙌  Thanks Nick!

NickEntin:
Haha I think I'm pretty confident in this now. Just trying to double check everything as I go since I don't have a lot of context around SPM. I appreciate all the investigation you've done here! 🙌 

As soon as this goes green, I'll merge the PR and enable the SPM Build job as a required check.

Thanks for another great addition, James! 🎉

Example/Podfile.lock
```diff
@@ -37,7 +37,7 @@ CHECKOUT OPTIONS:
     :git: https://github.com/square/Paralayout
 
 SPEC CHECKSUMS:
-  AccessibilitySnapshot: cb80a36813f8897825eada35ca5b09b210e472c3
+  AccessibilitySnapshot: 83a064dbe949ad93a8f24f916a558d5ac172b7d1
```
NickEntin, line 40: Oh, one more thing... ran `pod install` locally again and looks like the checksum changed since this was added.
```suggestion
  AccessibilitySnapshot: 4e1a6b6a16d945a4aca3a7459490fc6e1ccc867e
```

PR #50, opened 2020-12-09
Add support for snapshotting with Button Shapes enabled

NickEntin:
UIKit introduced a new API in iOS 14 to [check whether button shapes](https://developer.apple.com/documentation/uikit/uiaccessibility/3618943-buttonshapesenabled) are enabled. This adds a new snapshot method that mocks that setting being enabled.


NickEntin:
This currently works for buttons that check the properly / listen to the notification and manually update their appearance based on that. It doesn't work for standard buttons that use the system-defined button shape behavior though.

This is what the "enabled" snapshot should actually show:

simulator_screenshot_06FAFAB3-4320-4ED5-A9C9-6D6193497AB3

NickEntin:
Closing this for now. I'm not sure of a path forward with this to get all buttons working.

PR #43, opened 2020-11-04
Move iOS 12 CI build to GitHub Actions

NickEntin:
Travis CI [no longer has free macOS builds](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing), so it's time to migrate away. This moves our iOS 12 snapshot CI build over to GitHub Actions.

Issue #26, opened 2020-10-01
Tab bar descriptions are incorrect on iOS 14


NickEntin:
From testing on iOS 14.0.1, the actual description around tab bars differs slightly from what shows up in tests:

* The first item in the tab bar is prepended with "Tab Bar".
* In the case of fake tab bars (i.e. not using `UITabBarController`), some of the element descriptions no longer indicate the position in the tab bar (i.e. "x of y").

NickEntin:
This is still true on iOS 16

NickEntin:
Starting in iOS 17, it looks like in some cases it will read the badge as part of the description.

NickEntin:
> The first item in the tab bar is prepended with "Tab Bar".

This is still true on iOS 17 as well.

PR #52, opened 2021-01-02
Add iOS 14.2 to snapshot test matrix

NickEntin:


NickEntin:
Ugh, `SnapshotTestingTests.testInvertColors` is giving me different results when I step through it with a debugger vs. run it regularly... and neither are correct 🙃

NickEntin:
Now I'm getting the correct result when I step through it with a debugger (the snapshot I just added), but when it run it normally I still just get an all-black image.

efirestone (approved):

NickEntin:
I manually compared these snapshot against the demo app, and they all match with a few exceptions:
* Some of the data table descriptions are off (already tracked by #7)
* Some of the tab bar descriptions have changed slightly (already tracked by #26)
* It looks like there's a new automatic hint for the `.allowsDirectInteraction` trait to "Use the rotor to enable Direct Touch for this app." I'll file an issue to look into this.
* VoiceOver thinks some of my switches belong in the dining room... yeah, I have no idea

https://user-images.githubusercontent.com/1857549/103620229-0056cd80-4ee8-11eb-97a7-f971c1ddfb2c.mov

NickEntin:
I'm removing Travis CI as a required build since it isn't running the .com version right now (the .org version is passing, but will stop working soon). Hopefully we can migrate the iOS 12 tests over to GH Actions with #43.

Merging this now, then I'll add the iOS 14 CI build as a required PR check.

PR #69, opened 2021-08-26
Remove XCTest as a dependency

ra1028:
## Description

Removes XCTest as a dependency of the Core library.

## Motivation and Context

I'm using AccessibilitySnapshot through [PlaybookAccessibilitySnapshot](https://github.com/playbook-ui/accessibility-snapshot-ios). (Thanks for mentioning my library in the README before 😄 ). 
Recently, the increase in testing time to get the snapshots has become an issue. So, since my project uses Bazel as a build system, I am challenging the effort to use a remote cache for the reference snapshots.
As a start, Bazel doesn't allow output from test runs as artifacts, so I need to output from the app simulator runtime instead of unit-test.
I've already taken care of this in Playbook, but I want to change it so that we can run without XCTest in this library as well because PlaybookAccessibilitySnapshot depends on this library.
Apparently, the Core library hasn't actually depended on XCTest, but since it was declared as a dependency, I removed it.

### DIFF
--- a/AccessibilitySnapshot.podspec
+++ b/AccessibilitySnapshot.podspec
@@ -30,6 +30,8 @@ Pod::Spec.new do |s|
 
     ss.dependency 'AccessibilitySnapshot/Core'
     ss.dependency 'iOSSnapshotTestCase', '~> 6.0'
+    ss.frameworks = 'XCTest'
+    ss.weak_frameworks = 'XCTest'
   end
 
   s.subspec 'SnapshotTesting' do |ss|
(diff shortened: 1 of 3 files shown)


NickEntin:
Thanks for opening this! You're totally right, after we split the framework into subspecs, there's nothing in the `Core` spec that actually needs XCTest. This is a great update to make the core framework usable in more places.

Before I merge this, have you signed our [CLA](https://docs.google.com/forms/d/e/1FAIpQLSeRVQ35-gq2vdSxD1kdh7CJwRdjmUA0EZ9gRXaWYoUeKPZEQQ/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1)?

ra1028:
@NickEntin 
Thanks for taking a look at this. Signed!

NickEntin:
Awesome, thanks Ryo! Merging this in now and will plan on pushing a new release soon (hopefully this evening).

NickEntin (approved):

PR #72, opened 2021-09-08
Add support for using iOSSnapshotTestCase as the snapshot engine when installing via SPM

NickEntin:


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.4
```
NickEntin, line 1: This bumps the minimum required Xcode version for installing via SPM up to 12.5, which isn't ideal. This is required because iOSSnapshotTestCase [was bumped to 5.4 as well](https://github.com/uber/ios-snapshot-test-case/commit/511f302b587a72aed5022c03692e5e1ed2d985a3#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR1). @alanzeino is there a particular reason you need Swift 5.4?

NickEntin:
- [ ] Verify that the ObjC target still works now that it's accessing private methods in a different target.

Package.swift
alanzeino: I think only because we didn't have many other Xcode versions installed at the time to test and since we were newly introducing SPM support it made sense to start from the latest and go from there.

Testing 5.3 here https://github.com/uber/ios-snapshot-test-case/pull/162

alanzeino: Done. Do you need a new version or can you use a commit ref for now?

NickEntin: Thanks! I think a new version would be preferable. This would probably be a patch release?

NickEntin: I can test with a commit ref for now, so no rush on getting that released. But will probably want a release before I merge this.

NickEntin:
Closing this PR in favor of #97 (same change, just not using a fork to make stacking work)

PR #74, opened 2021-09-27
Disable rendering large images in monochrome on iOS 13

### LINKED ISSUE #73: Monochrome rendering doesn't work on large images in iOS 13
`CIImage` renders a blank image when applying a `CIFilter` to larger images, which results in blank accessibility content in the accessibility snapshot images. This appears to be iOS 13 specific and does not apply to iOS 12 or 14.

NickEntin:
Resolves #73


Sources/AccessibilitySnapshot/SnapshotTesting/SnapshotTesting+Accessibility.swift
```diff
@@ -67,7 +67,16 @@ extension Snapshotting where Value == UIView, Format == UIImage {
                 containerView.center = window.center
                 window.addSubview(containerView)
 
-                containerView.parseAccessibility(useMonochromeSnapshot: useMonochromeSnapshot)
+                do {
+                    try containerView.parseAccessibility(useMonochromeSnapshot: useMonochromeSnapshot)
+                } catch AccessibilitySnapshotView.RenderError.imageExceedsMaximumSize {
+                    fatalError(
+                        "View is too large to render monochrome snapshot. Try setting useMonochromeSnapshot to false."
```
efirestone, line 74: Small recommendations:
- Make the error itself `StringConvertible` and move this error text there. The code producing the error has better context, and can more safely explain it. It also means you can print the error description both here and in the FBSnapshot version without repeating it.
- Add some additional context to the error message about iOS 13. You can either set `useMonochromeSnapshot: false`, OR you can use an operating system other than iOS 13. Worth mentioning that this is fixed on iOS 14 and later.

efirestone (approved):

Sources/AccessibilitySnapshot/SnapshotTesting/SnapshotTesting+Accessibility.swift
NickEntin: > The code producing the error has better context, and can more safely explain it.

Technically the error should get converted at various points to add more context (the generic image utility produces an error saying the `image` is too large, the snapshot container view converts that to an error saying that the `containedView` was too large, and then the snapshot method converts it to an error saying that the view under test was too large). The generic image utility is private though, so I'm not concerned about that abstraction layer. So really there are two levels of surfacing the error.

I agree the repetition isn't great though. I updated this to provide a better error at the `AccessibilitySnapshotView` level (in the `Core` spec). I also consolidated the error message logic between the Swift and Objective-C methods for iOSSnapshotTestCase into a factory. Due to how the framework boundaries are set up, there's not a great way to share this with the SnapshotTesting version, but I think a small amount of repetition here is probably acceptable. I also updated the error messages to specify updating to iOS 14 is a known fix.

PR #76, opened 2021-09-29
Tile and stitch images when snapshotting large views

NickEntin:


NickEntin:
```
❌  /Users/runner/work/AccessibilitySnapshot/AccessibilitySnapshot/Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilitySnapshotView.swift:852:34: use of unresolved identifier 'Self'; did you mean 'self'?

            tileRect.origin.y += Self.tileSideLength
              ^
```

Oh Xcode 10... I think since we're gonna need a major release anyway, it's time to bump the minimum to Xcode 12.

Sources/AccessibilitySnapshot/iOSSnapshotTestCase/ErrorMessageFactory.swift
```diff
@@ -29,6 +29,11 @@ internal enum ErrorMessageFactory {
                 View is too large to render monochrome snapshot. Try setting useMonochromeSnapshot to false or use a \
                 different iOS version. In particular, this is known to fail on iOS 13, but was fixed in iOS 14.
                 """
+        case AccessibilitySnapshotView.Error.containedViewHasUnsupportedTransform:
```
efirestone, line 32: Same comment as the other PR...we should move these error message strings into a centralized spot. Non-blocking for sure.

efirestone (approved):

Sources/AccessibilitySnapshot/iOSSnapshotTestCase/ErrorMessageFactory.swift
NickEntin: Same issue here with the current module boundaries making that tricky. It's split up here the same way (shared for the Swift and Objective-C implementations of iOSSnapshotTestCase, but separate for the SnapshotTesting implementation), but I'll think about whether there's a way to combine those.

PR #77, opened 2021-09-29
Bump minimum Xcode version to 12

NickEntin:


Scripts/github/prepare-simulators.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -l
```
NickEntin, line 1: This script follows the format used in Paralayout and Valet.

efirestone (approved):

PR #89, opened 2022-05-04
Handle -[UIApplication keyWindow] deprecation

marcelofabri:
This was deprecated on iOS 13. It’s not currently a warning because the library’s deployment target is iOS 12, but it’ll be soon - plus it can be a warning if integrating the library via submodule or using the source code directly.

### DIFF
--- a/Sources/AccessibilitySnapshot/SnapshotTesting/SnapshotTesting+Accessibility.swift
+++ b/Sources/AccessibilitySnapshot/SnapshotTesting/SnapshotTesting+Accessibility.swift
@@ -122,7 +122,18 @@ extension Snapshotting where Value == UIView, Format == UIImage {
             let requiresWindow = (view.window == nil && !(view is UIWindow))
 
             if requiresWindow {
-                let window = UIApplication.shared.keyWindow ?? UIWindow(frame: UIScreen.main.bounds)
+                let keyWindow: UIWindow? = {
+                    if #available(iOS 13.0, *) {
+                        return UIApplication
+                            .shared
+                            .connectedScenes
+                            .flatMap { ($0 as? UIWindowScene)?.windows ?? [] }
+                            .first { $0.isKeyWindow }
+                    } else {
+                        return UIApplication.shared.keyWindow
+                    }
+                }()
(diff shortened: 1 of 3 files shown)


NickEntin:
Thanks for submitting this @marcelofabri!

@jhneves Mind taking a look at this one? I don't think this will cause any issues, but want to make sure we aren't missing anything around apps that don't use scenes.

Sources/AccessibilitySnapshot/SnapshotTesting/SnapshotTesting+Accessibility.swift
```diff
@@ -122,7 +122,18 @@ extension Snapshotting where Value == UIView, Format == UIImage {
...
+                        return UIApplication
+                            .shared
+                            .connectedScenes
+                            .flatMap { ($0 as? UIWindowScene)?.windows ?? [] }
+                            .first { $0.isKeyWindow }
+                    } else {
+                        return UIApplication.shared.keyWindow
+                    }
+                }()
+                let window = keyWindow ?? UIWindow(frame: UIScreen.main.bounds)
```
jhneves, line 136: Thanks @marcelofabri for the PR 🙌 

it might be worth moving the keyWindow grabbing function somewhere (e.g. an extension of UIApplication, so we only write it once).

marcelofabri: the main issue is that this extension would need to be declared `public` inside `AccessibilitySnapshot/Core` and I don't think this should be exposed to outside consumers

jhneves: Perhaps a new path [here](https://github.com/cashapp/AccessibilitySnapshot/tree/master/Sources/AccessibilitySnapshot) to include shared files ? Then we could update the podspec/swift package to include that directly as a source to keep it internal.

jhneves: no problem. I will take care of it. Thanks @marcelofabri !

NickEntin:
This was fixed in #116. Thanks for this PR @marcelofabri!

PR #91, opened 2022-06-06
Fix and add tests for iOS 15 devices

luispadron:
## Summary

The `fishhook.c` code in the repository had started crashing for iOS 15 devices. I copied the contents from the `.c` file from the fix in this PR: https://github.com/facebook/fishhook/pull/87

I also attempted to add tests for iOS 15, not sure if I added them correctly to CI but I generated the snapshots and tested locally with iOS 15.2 iPhone 13 Pro.


luispadron:
@NickEntin @jhneves 🙏🏼

jhneves:
diff appears fine to me, but definitely worth having @NickEntin 👀  on this one. The CI seems like does not have xcode 13.2 though @luispadron ?

luispadron:
@jhneves Thanks for checking! Yeah I'm not entirely sure how to use new Xcode with GitHub actions, I'll defer to @NickEntin here 🙏🏼

PR #97, opened 2022-08-16
Add support for using iOSSnapshotTestCase as the snapshot engine when installing via SPM

NickEntin:


NickEntin:
- [ ] Verify that the ObjC target still works now that it's accessing private methods in a different target.

alanzeino:
Let me know if you need me to do anything upstream in our repo!

Sources/AccessibilitySnapshot/iOSSnapshotTestCase/ObjC/include/FBSnapshotTestCase_Accessibility.h
```diff
@@ -14,7 +14,11 @@
 //  limitations under the License.
 //
 
-#import <FBSnapshotTestCase/FBSnapshotTestCase.h>
+//#if SWIFT_PACKAGE
+@import iOSSnapshotTestCaseCore;
+//#else
+//#import <iOSSnapshotTestCaseCore/FBSnapshotTestCase.h>
+//#endif
```
NickEntin, line 21: This is the problem. Ideally this should be
```c
#if SWIFT_PACKAGE
@import iOSSnapshotTestCaseCore;
#else
#import <FBSnapshotTestCase/FBSnapshotTestCase.h>
#endif
```
(I think the `@import` should actually be `#import` with a specific file, but that wouldn't compile 🤷‍♂️)

The issue is that `SWIFT_PACKAGE` isn't available in public header files as far as I can tell, so there's no way to distinguish whether it's being built with SPM or CP.

@alanzeino any idea how to work around this?

NickEntin: Nevermind, turns out this import isn't actually needed at all 🤦‍♂️

AliceAponasko (approved):

Sources/AccessibilitySnapshot/iOSSnapshotTestCase/ObjC/FBSnapshotTestCase_Accessibility.m
```diff
@@ -0,0 +1,18 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+// This file is intentionally left blank. SPM requires the target to have an implementation file to recognize that it is
+// an Objective-C target.
```
jhneves, line 18: TIL

Issue #63, opened 2021-05-25
Add precision and tolerance apis


fruitcoder:
The missing `tolerance` was mentioned here #15 but no PR was added yet. I'll leave this here while I'm experimenting with the implementation to fix failing snapshots on my CI.

mattprowse-xero:
I'm in the process of adding AccessibilitySnapshot to a project I'm working on, and this is our last remaining blocker. I noticed #64 from a bit over a year ago: is there anything I can do to help get this over the line? More than happy to create a new PR if needed.

fruitcoder:
We're still using our fork. Would be great to get this merged ir any info what would contribute to it being merged

NickEntin:
@fruitcoder Does using per-pixel tolerance alone (without overall tolerance) fix your CI issues? I think that one is a bit safer than overall tolerance.

fruitcoder:
Are you talking about the `perceptualPrecision` that's mentioned in the new swift-snapshit-testing 1.10.0 release or was this something I could have used before. If so, can you point me in the right direction? 😅

NickEntin:
Ahh you're using SnapshotTesting. Per-pixel tolerance is only supported out of box in iOSSnapshotTestCase right now (see.

Perceptual precision is very similar to per-pixel precision, just using a different definition for how "similar" two pixels are. It potentially could be more reliable than standard per-pixel precision, but I haven't had a chance to run it through our test suite yet.

NickEntin:
Hey @fruitcoder, just following up here - did you get a chance to test whether per-pixel/perceptual precision is sufficient to fix your CI issues?

fruitcoder:
Hey @NickEntin! Is there anything I can test here? I'm using snapshot testing and still can't find a way to get an image strategy that uses any kind of tolerance 🤔

NickEntin:
@fruitcoder Try [this branch](https://github.com/cashapp/AccessibilitySnapshot/compare/master...entin/perceptual-precision?expand=1). That should give you a parameter for `.accessibilityImage(perceptualPrecision: ...)`. Try something like `0.995` to start.

albsala-backmarket:
Hi @NickEntin! It seems we removed the CocoaPods/Bazel blocker from #143. Could we move forward with this topic? It's really a blocker for using the library on CI 😿.
Thank you!

DimitarNestorov:
[...]

PR #104, opened 2022-09-08
Update workflow with macOS 11

alcivanio:
Update workflow with macOS 11


alcivanio:
This patch is to update the os version our ci will run. A few other files needed to be update, things such as devices sizes, some frameworks versions where also updated.

gonzaloperretti:
Hi, any news regarding this?

Example/SnapshotTests/LocalTestsConstants.swift
```diff
@@ -0,0 +1,41 @@
...
+import Foundation
+import UIKit
+
+public class LocalTestsConstants: NSObject {
+    
+    /**
+     This float ranges from 0 to 1 and will be used to calculate how precise a snapshot test will be,
+     where 1 means all pixels should match, and 0 means that none of them need to match.
+     */
+    static public let testPrecision: CGFloat = 0.9
```
NickEntin, line 26: 90% seems pretty low... does it need to be this low to pass, or can we be stricter?

Sources/AccessibilitySnapshot/iOSSnapshotTestCase/ObjC/include/FBSnapshotTestCase_Accessibility.h
```diff
@@ -14,47 +14,47 @@
 //  limitations under the License.
 //
 
-#define SnapshotVerifyAccessibility(view__, identifier__)\
+#define SnapshotVerifyAccessibility(view__, identifier__, perPixelTolerance__)\
```
NickEntin, line 17: We should keep the basic two argument macro and add a new one for tolerance. I think many consumers will still want to non-tolerance version.

Example/SnapshotTests/LocalTestsConstants.swift
alcivanio: That's a good catch. It was a type, but I haven't tested with other numbers. Let's try 98% to see how it goes.

NickEntin:
This is missing a bunch of file deletions... we've added reference images for 14.5 and 13.7, but haven't deleted the old ones for 14.2 and 13.3.

I spent way too long trying to figure out why Git wasn't able to figure out a lot of these were renames before I realized that was the issue. In any case, ended up with #107.

Scripts/build.swift
```diff
@@ -46,18 +46,18 @@ enum TaskError: Error {
 }
 
 enum Platform: String, CustomStringConvertible {
+    case iOS_15
```
NickEntin, line 49: Let's not add 15 yet, since we aren't actually running it here. Better to have one update that adds iOS 15 coverage than slowly add it over multiple PRs and lose track of what's supported and what's not.

NickEntin:
At a higher level, this PR is getting quite large. From what I can tell, there's at least X distinct tasks here:

1. Update iOS 13 snapshots from iOS 13.3 to 13.7.
2. Update iOS 14 snapshots from iOS 14.2 to 14.5.
3. Introduce comparison tolerances.
4. Update the CI runner from `macOS-10.15` to `macOS-11` and associated gem/action dependencies.
5. Change sample project dependencies to use SnapshotTesting 1.9 instead of 1.7.2.

Unfortunately the runners don't overlap nicely, so 2 and 4 are tied together, but I think it would be a lot cleaner to move 1 and 3 (and maybe 5) to separate PRs. Especially 3 since it's a bit of a complicated topic (see #63).

PR #27, opened 2020-10-02
Bump minimum supported iOS version to 12.0

NickEntin:
Resolves #25.

PR #113, opened 2022-12-19
Remove iOS 11 snapshots from example

eliperkins:
These were leftover from #27 and are no longer used.


NickEntin:
🤦‍♂️  Oops, good catch!

NickEntin (approved):

PR #116, opened 2023-01-19
Bump Cocoapods minimum deployment target to iOS 13

n8chur:
Bumps the Cocoapods iOS minimum deployment target of this framework to be in line with the iOS deployment target specified in the Swift Package.

- Bump minimum deployment target to iOS 13 in podspec
- Remove iOS 12-specific code from sourcecode
- Bump deployment target of example project
- Remove iOS 12-specific code from example
- Update README with iOS 13 bump information
- Use scene APIs to determine key window since `UIApplication.shared.keyWindow` was deprecated


Sources/AccessibilitySnapshot/Core/Swift/Classes/UIAccessibility+SnapshotAdditions.swift
```diff
@@ -22,7 +22,7 @@ extension NSObject {
     func accessibilityDescription(context: AccessibilityHierarchyParser.Context?) -> (description: String, hint: String?) {
         var accessibilityDescription = accessibilityLabelOverride(for: context) ?? accessibilityLabel ?? ""
 
-        var hintDescription = hidesAccessibilityHint() ? nil : accessibilityHint?.nonEmpty()
```
NickEntin, line 25: I think we still need this since there are reasons that could be true other than the OS version.

n8chur: I can restore it, but if I'm understanding the logic correctly, the implementation of that check would now just return `false`, right?
This:
```swift
    private func hidesAccessibilityHint() -> Bool {
        if NSObject.majorOSVersion >= 13 {
            return false
        } else {
            return accessibilityTraits.contains(.tabBarItem)
        }
    }
```
becomes:
```swift
    private func hidesAccessibilityHint() -> Bool {
        false
    }
```
if we're always building for iOS 13 and above.

NickEntin: Oh sorry, GitHub was showing the diff weird and it looked like there was more in that method. You're right.

NickEntin (approved):

PR #125, opened 2023-04-18
[DNM] Debugging CI

NickEntin:


eliperkins:
@NickEntin hmmmm, I don't have any specific context! From looking at the raw logs, it seems like builds are getting hung up on a codesigning-ish step, which leads me to think that this might be an `xcodebuild` problem, rather than an Actions one. The last message logged is consistently:

```
Capabilities for Signing & Capabilities may not function correctly because its entitlements use a placeholder team ID. To resolve this, select a development team in the AccessibilitySnapshotDemo editor. (in target 'AccessibilitySnapshotDemo' from project 'AccessibilitySnapshot')
```


If it were Actions runners that were having issues, cancelling the build would result in other disconnect, termination or timeout messages.

I wonder if `xcodebuild` itself is hanging here for some reason. It's a bit baffling why only the iOS 14 build would be affected though. What are some other diagnostics we can collect from `xcodebuild` here to diagnose further?

NickEntin:
I wonder if this is the same underlying issue as actions/runner-images#7687. It looks like those are getting hung up around the code sign step too, and similarly on macOS 11 and Xcode 12.

PR #149, opened 2023-08-16
Move SnapshotTesting tests to SPM test target

NickEntin:
This moves the SnapshotTesting tests out of the demo app, which unblocks dropping CP support for the SnapshotTesting extension, which unblocks bumping the SnapshotTesting dependency to 1.10, which unblocks #143.


NickEntin:
The tests currently fail on:

> Accessibility snapshot tests cannot be run in a test target without a host application

Unfortunately it doesn't look like there's a way to get SPM test targets to run inside a host application, so I don't think this approach is going to work. 😞

Issue #59, opened 2021-04-27
Snapshots do not reflect the use of accessibilityUserInputLabels


annahayhurst:
As of iOS 13, the [`accessibilityUserInputLabels`](https://developer.apple.com/documentation/objectivec/nsobject/3197989-accessibilityuserinputlabels) array can be used as an alternative to a singular `accessibilityLabel` on an accessibility element. As per the dev docs, iOS respects this array as first priority and falls back to the standard label if it returns as empty or nil.

In our app, we have a button that sets this property with an array of two strings if the property is available. When AX snapshots are taken, they do not load the label(s) from the array. I would expect the snapshot to either reflect the first element (in our case, "Fast Forward") or to show some representation of the array, but it instead shows the underlying button name `forward15` as if we never set any AX info.

The app itself handles the array fine (we use it for voice input) and if I set the usual `accessibilityLabel` the snapshot shows that label. Tests are running on iOS 14.3.

NickEntin:
Great catch, Anna! Thanks for opening this issue. We should definitely add support for `accessibilityUserInputLabels`. I think the right behavior here is to mirror the way VoiceOver would describe the element based on the given labels.

Are you interested in opening a PR to update this behavior?

fruitcoder:
Isn't `accessibilityUserInputLabels` currently only used for VoiceControl? We use it as well and VoiceOver still just reads the a11yLabel on views/accessibility elements that have both properties set. I think it would make sense to be able to choose which labels to use and to provide a similar display mode like "accessibilityLabelDisplayMode" where you can choose `whenOverriden`/`a11Label`/`a11yInputLabels`. `whenOverriden` would use the input labels when `accessibilityUserInputLabels` is set and contains more than one item or just one item that is different from the a11y label.

NickEntin:
Looks like you're right, Alex. I did some tests on iOS 14.3 and VoiceOver doesn't seem to change any behavior based on the `accessibilityUserInputLabels`, regardless of whether the `accessibilityLabel` is set or not.

Rather than replacing the `accessibilityLabel` in the snapshot, I think we should instead show any user input labels in addition to the VoiceOver description. I'll mock up some designs for this, but I think the key will be making it clear that these are individual items that can be used to refer to the element, rather than used to make a composite description.

PR #150, opened 2023-08-17
Add pills view for accessibility input labels

meherkasam:
[`accessibilityUserInputLabels`](https://developer.apple.com/documentation/objectivec/nsobject/3197989-accessibilityuserinputlabels) allow the developer to provide an improved experience for a Voice Control user to refer to a specific element when the default label (provided by `accessibilityLabel`) is not suitable for voice input. This PR addresses the feature request raised in [this](https://github.com/cashapp/AccessibilitySnapshot/issues/59) issue.

Here's an example of what the legend would look like:
image


meherkasam:
@NickEntin As it's currently written, it will be a breaking change for consumers. Would you recommend adding a parameter to `AccessibilitySnapshotView` such as `showInputs: Bool = false`?

NickEntin:
> @NickEntin As it's currently written, it will be a breaking change for consumers. Would you recommend adding a parameter to `AccessibilitySnapshotView` such as `showInputs: Bool = false`?

No, I think it's fine for this to be a breaking change. I'm not opposed to making showing the inputs optional, but if so we should default to enabled.

NickEntin:
@meherkasam This is looking great! What do you think about this for the pill style instead?

testActivationPointEnabled_16_4_393x852@3x

No border, increase horizontal padding to 6 pt, background color 80% gray, text color 20% gray.

This looks a lot closer to how the pills actually look with Voice Control (while still in the style of the legend), so I think it's a nice little visual indicator to where these apply in the UI.

meherkasam:
@NickEntin Thanks for the feedback. I've updated my PR and the description with the latest screenshot. PTAL and let me know if it looks okay.

Example/AccessibilitySnapshot/ActivationPointViewController.swift
NickEntin, line 1: Let's add a new sample VC for this that hits an example of:

* no labels
* one label
* multiple lines of labels
* a really long label

NickEntin (commented):
This is looking great! Thanks for putting this together.

meherkasam:
@NickEntin I've incorporated all of your suggestions. PTAL now.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilitySnapshotView+PillsView.swift
```diff
@@ -0,0 +1,149 @@
...
+    class PillsView: UIView {
+        private let pills: [PillView]
+        
+        init(titles: [String], color: UIColor) {
+            pills = titles.map { PillView(title: $0, color: color) }
+            
+            super.init(frame: .zero)
+            
+            pills.forEach {
+                $0.sizeToFit()
```
NickEntin, line 91: This isn't necessary since we resize in `layoutSubviews` anyway

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilitySnapshotView+PillsView.swift
```diff
@@ -0,0 +1,149 @@
...
+            label.frame = CGRect(
+                x: bounds.minX + Metrics.pillHorizontalPadding,
+                y: bounds.minY + Metrics.pillVerticalPadding,
+                width: bounds.maxX - 2 * Metrics.pillHorizontalPadding,
+                height: bounds.maxY - 2 * Metrics.pillVerticalPadding
+            )
+        }
+    }
+    
+    class PillsView: UIView {
```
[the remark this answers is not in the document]
meherkasam: I get the following warning in Xcode when I add this:
```
'internal' modifier is redundant for class declared in an internal extension
```

PR #167, opened 2023-10-18
Add snapshotting methods to snapshot SwiftUI views directly

### LINKED ISSUE #37: Add SwiftUI snapshot support
SnapshotTesting supports SwiftUI views so should be a case of adding the relevant extension to our spec code. This'll avoid our users from having to first wrap the view in a UIHostingController simplifying it's usage.

NickEntin:
This simplifies the process for snapshotting SwiftUI views by avoiding the need to wrap the view in a `UIHostingController`.

Resolves #37


Example/SnapshotTests/SnapshotTestingTests.swift
```diff
@@ -28,15 +28,19 @@ final class SnapshotTestingTests: XCTestCase {
     #if swift(>=5.1) && canImport(SwiftUI)
```
meherkasam, line 28: Do you want to remove this macro check from this test since you removed it from the source file?

meherkasam (approved):

Example/SnapshotTests/SnapshotTestingTests.swift
NickEntin: Hmm yeah I think we can probably remove it now. When we added it Swift 5.1 was new, but now all of our supported Xcode versions are >5.1.

Issue #129, opened 2023-05-24
Incorrect sort order


yonaskolb:
The order of parsed accessibility markers is in the wrong order. See screenshot, where the section headings are parsed as being at the bottom of the view. This is a SwiftUI view with no animations. Is there a workaround for this?
The order is important to me as I use the accessibility markers to create a markdown representation of the view.

![testExample MainA11y](https://github.com/cashapp/AccessibilitySnapshot/assets/2393781/71381c0d-d0ae-47ec-a023-b8aba3f57038)

NickEntin:
Hi @yonaskolb, thanks for reporting this! Can you provide a code sample that reproduces this issue? Additionally, what iOS version are you testing on?

yonaskolb:
Thanks for the quick reply @NickEntin. Example project showing the issue attached. I'm using iOS 16.4 in Xcode 14.3
[SnapshotExample.zip](https://github.com/cashapp/AccessibilitySnapshot/files/11551001/SnapshotExample.zip)
![testExample Form](https://github.com/cashapp/AccessibilitySnapshot/assets/2393781/73e14043-df47-492d-ad20-55ddc25037e7)

yonaskolb:
The example also highlights a different issue with the rendering of forms, which there is an unresolved issue in SnapshotTesting for https://github.com/pointfreeco/swift-snapshot-testing/issues/667

yonaskolb:
Any updates on this one?

NickEntin:
I haven't had a chance to test on Xcode 14.3 yet, but on Xcode 13.3 it seems to be behaving correctly. I'll dig in further this week.

NickEntin:
Still working on this one. There was definitely a change in iOS 16 that we're not currently handling properly. Looking at sample view you provided, in iOS 15 it's backed by a table view and in iOS 16 it's backed by a collection view.

<img width="1357" alt="iOS 15" src="https://github.com/cashapp/AccessibilitySnapshot/assets/1857549/b3453b5c-ee9e-46bd-a9b4-af6eeab55e1d">
<img width="1288" alt="iOS 16" src="https://github.com/cashapp/AccessibilitySnapshot/assets/1857549/fbd7087b-ab5f-44fd-86de-0886fdaa2076">

It looks like with collection views it's using an accessibility container with a single element view (that's not an accessibility element) which then has subviews that are themselves accessibility containers each with a single element. Somewhere in this nesting we don't order things the same as VoiceOver.

[...]

PR #169, opened 2023-11-05
Fix incorrect sort order of NavigationStack

### LINKED ISSUE #168: Incorrect sort order when using NavigationStack
I've found that when using a `NavigationStack` in a SwiftUI view accessibility snapshot puts the contents of the navigation bar below the contents of the view that is contained in the `NavigationStack`. That is a lot of words that may or may not be helpful so I'll share some snapshots:

|NavigationView|NavigationStack|
|-|-|
|testNavigationView 1|testNavigationStack 1|

Here is the code I used to generate those snapshots:

```swift
import AccessibilitySnapshot
import SnapshotTesting
import SwiftUI
import XCTest

class NavigationExampleSnapshotTests: XCTestCase {
  func testNavigationView() {
    let view = NavigationView {
      Text("Text inside a NavigationView")
        .navigationTitle("Navigation View")
        .toolbar {
          ToolbarItem {
            Button {
              // no-op
            } label: {
              Text("Add")
            }
          }
        }
    }

    let hostingController = UIHostingController(rootView: view)
    hostingController.view.frame = UIScreen.main.bounds

    assertSnapshot(
      of: hostingController,
      as: .accessibilityImage(drawHierarchyInKeyWindow: true)
    )
  }

  func testNavigationStack() {
    let view = NavigationStack {
      Text("Text inside a NavigationStack")
        .navigationTitle("Navigation Stack")
        .toolbar {
          ToolbarItem {
            Button {
              // no-op
            } label: {
              Text("Add")
            }
          }
        }
    }

    let hostingController = UIHostingController(rootView: view)
    hostingController.view.frame = UIScreen.main.bounds

    assertSnapshot(
      of: hostingController,
      as: .accessibilityImage(drawHierarchyInKeyWindow: true)
    )
  }
}
```

I've noticed this issue when running Xcode 15 in an iOS 17 simulator. I have not tested previous simulators but I think the issue is related to `NavigationStack` and not to the simulator.

DavidBrunow:
Fixes #168.


Example/SnapshotTests/__Snapshots__/SnapshotTestingTests/testNavigationStack.393x852-16-4-3x.png
DavidBrunow, line 1: I'm on an Apple Silicon Mac so these snapshots might not be generated as needed.

Example/AccessibilitySnapshot/SwiftUIView+EmbedInHostingController.swift
```diff
@@ -0,0 +1,30 @@
+//
```
DavidBrunow, line 1: I know y'all just got rid of this, but I brought it back because the navigation buttons were not being rendered when using only the SwiftUI views.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -601,9 +601,14 @@ private extension NSObject {
                     )
                 )
             }
+
+            let explicitlyOrdered = accessibilityElements.contains {
+                "\(type(of: $0))" != "UILayoutContainerView"
```
DavidBrunow, line 606: I don't love this fix but from my very small set of tests it works. Next week I'm going to check against our tests at work to see if anything breaks with this change.

Example/SnapshotTests/SnapshotTestingTests.swift
```diff
@@ -138,6 +138,32 @@ final class SnapshotTestingTests: XCTestCase {
         assertSnapshot(matching: viewController, as: .imageWithSmartInvert, named: nameForDevice())
     }
 
+    @available(iOS 16.0, *)
```
DavidBrunow, line 141: If I remember correctly this doesn't do quite enough to make the tests pass and I'll need to add an `XCTSkip`.

Example/AccessibilitySnapshot/SwiftUIView+EmbedInHostingController.swift
NickEntin: Hmm, they weren't being rendered at all? The only thing that should be different between these is how the view is sized (setting the `frame` vs. `bounds`). Might affect the safe area insets, but everything should still be rendered.

DavidBrunow: I'm sorry, I misspoke. The button is getting rendered but it is not included as an accessibility element

testNavigationStack 393x852-16-4-3x

Changing this code from `bounds` to `frame` resolves the issue: https://github.com/cashapp/AccessibilitySnapshot/blob/master/Sources/AccessibilitySnapshot/SnapshotTesting/SnapshotTesting%2BSwiftUI.swift#L68

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
DavidBrunow: I've checked our tests at work and see no regression from this change.

DavidBrunow:
@NickEntin any feedback on this pull request? I think it is ready to come out of draft status but I want to make sure y'all are good with it.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -601,9 +601,19 @@ private extension NSObject {
                     )
                 )
             }
+
+            let explicitOrderingExceptionList = [
+                "UILayoutContainerView",
+                "UpdateCoalescingCollectionView",
+            ]
```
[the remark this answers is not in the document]
DavidBrunow: Update with what I've found so far:

Both the `NavigationView` and `NavigationStack` have a `UILayoutContainerView` in their view hierarchy -- in general both view hierarchies look very similar. The key difference between the two is that the `UILayoutContainerView` in the `NavigationView` has `shouldGroupAccessibilityChildren` set to `true` while in the `NavigationStack` it is set to `false`. 

This is key to the problem because the `shouldGroupAccessiblityChildren` property is what is used to determine whether a group of accessibility elements has `explicitlyOrdered` set to `false`: https://github.com/cashapp/AccessibilitySnapshot/blob/master/Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift#L631

To me this points to what you mentioned: a larger issue around how the library is matching VoiceOver's treatment of accessibility containers. In specific I think we need improvements to how `explicitlyOrdered` is determined, or maybe how it is used to create the sorted nodes, but I don't have suggestions at this time.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -601,9 +601,28 @@ private extension NSObject {
...
+
+            let explicitlyOrdered = accessibilityElements.contains {
+                explicitOrderingExceptionList.contains("\(type(of: $0))") == false
+            }
+            let shouldBeExplicitlyOrdered = accessibilityHierarchyOfElements.contains { node in
```
DavidBrunow, line 613: Here is another approach to determining whether a group should be explicitly ordered. I need to think it through a bit more but I think this matches how VoiceOver works since the ordering is determined by the order of the `accessibilityElements` array.

yonaskolb:
Would love to see this land at some point. Seems it may fix https://github.com/cashapp/AccessibilitySnapshot/issues/129 as well?

DavidBrunow:
> Just tested and this does indeed fix #129

I need to pick this back up again, will try to do so soon.

yonaskolb:
Any updates on this one, especially after #244 was merged?

PR #170, opened 2023-11-10
Check UISwitch state using combination of trait and value

NickEntin:


NickEntin:
- [ ] Add test cases showing this behavior
- [ ] Check if the outputted values are localized by numeric script
- [ ] Check what happens when the accessibilityValue is a non-0 or 1 numeric value

NickEntin:
> Check what happens when the accessibilityValue is a non-0 or 1 numeric value

From what I can tell there are three special cased values:
* `0` => "off"
* `1` => "on"
* `2` => "mixed"

I'm not sure what the last one is used for, but simple enough to support with the same pattern.

NickEntin:
> Check if the outputted values are localized by numeric script

Nope, looks like it's always `0` and `1`

NickEntin:
- [ ] Update iOS 13 reference image from CI
- [ ] Update localized descriptions for German
- [ ] Update localized descriptions for Russian

NickEntin:
> > Check what happens when the accessibilityValue is a non-0 or 1 numeric value
> 
> From what I can tell there are three special cased values:
> 
> * `0` => "off"
> * `1` => "on"
> * `2` => "mixed"
> 
> I'm not sure what the last one is used for, but simple enough to support with the same pattern.

Oh, VoiceOver probably shares logic with macOS, which has [tri-state controls](https://developer.apple.com/documentation/appkit/nscontrolstatevalue/) (e.g. checkboxes).

olkosenko (approved):

meherkasam-square (approved):

meherkasam (approved):

PR #173, opened 2023-11-28
Support Accessibility Custom Content

RoyalPineapple:
Introduced in iOS 14 [Accessibility Custom Content](https://developer.apple.com/documentation/accessibility/customized_accessibility_content) allows the inclusion of arbitrary accessibility content defined using pairs of label and value. 
As a default this additional data is exposed under a "Custom Content" rotor entry, but can also be included in the element's VoiceOver utterance by setting the content's `importance` instance variable to `.high` or by adjusting verbosity in VoiceOver settings. 

This PR adds support for displaying custom content similar to how we handle custom actions.


Example/AccessibilitySnapshot/SwiftUIView.swift
```diff
@@ -77,8 +77,27 @@ struct SwiftUIView: View {
                 .accessibility(label: Text("Label"))
                 .accessibility(value: Text("Value"))
                 .accessibility(hint: Text("Hint"))
-
-            Spacer()
```
RoyalPineapple, line 81: This removal of the spacer is notable in that it causes changes in all snapshots of this view by centering the contents however its inclusion causes the containing VStack to exceed 10 elements.

Example/AccessibilitySnapshot/SwiftUIView.swift
```diff
@@ -77,8 +77,27 @@ struct SwiftUIView: View {
...
+            if #available(iOS 14.0, *) {
+                // View with label, value, hint, and Custom Actions.
+                Circle()
+                    .accessibility(label: Text("Label"))
+                    .accessibility(value: Text("Value"))
+                    .accessibility(hint: Text("Hint"))
+                    .accessibilityAction(named: "Custom") {}
+            }
+// Swift 5.5 ships with Xcode 13 and the iOS 15 SDK.
+#if swift(>=5.5)
```
NickEntin, line 90: #197 should fix this so you don't need the Swift version checks for 5.4 or 5.5

Example/AccessibilitySnapshot/SwiftUIView.swift
NickEntin: Ugh arbitrary SwiftUI limitations...

Example/AccessibilitySnapshot/SwiftUIView.swift
RoyalPineapple: Now only needing a single check in the accessibilityParser. Much nicer. Thanks

Example/AccessibilitySnapshot/SwiftUIView.swift
meherkasam: Can we not use `Group` here to group the circles together and retain the spacer?

Example/AccessibilitySnapshot/SwiftUIView.swift
```diff
@@ -77,8 +77,25 @@ struct SwiftUIView: View {
...
+                // View with label, value, hint, and Custom Actions.
+                Circle()
+                    .accessibility(label: Text("Label"))
+                    .accessibility(value: Text("Value"))
+                    .accessibility(hint: Text("Hint"))
+                    .accessibilityAction(named: "Custom") {}
+            }
 
-            Spacer()
+            if #available(iOS 15.0, *) {
```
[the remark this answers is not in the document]
RoyalPineapple: Yea, it was added to UIKit in 14 and then made it to SwiftUI in 15.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilitySnapshotView.swift
```diff
@@ -283,6 +283,22 @@ private extension AccessibilitySnapshotView {
                 )
             }()
             
+            
+            // If our description and hint are both empty, and we dont have custom actions, but we do have custom content, we'll use the description label
+            // to show the "More Content Available" text, since this makes our layout simpler when we align to the marker.
+            let showCustomContentInDescription = (marker.description.isEmpty && !showActionsAvailableInDescription && !marker.customContent.isEmpty)
```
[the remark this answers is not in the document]
RoyalPineapple: Just tested this and yes VoiceOver reads "More Actions Available" and "More Content Available" if the element has no other accessibility properties set. 

In the case where an element has both actions and content the actions announcement comes first.

Example/AccessibilitySnapshot/AccessibilityCustomContentViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+            accessibilityHint = includeHint ? "Hint" : nil
+        }
+
+        @available(*, unavailable)
+        required init?(coder: NSCoder) {
+            fatalError("init(coder:) has not been implemented")
+        }
+
+        // MARK: - UIAccessibility
+        var accessibilityCustomContent: [AXCustomContent]! = {
```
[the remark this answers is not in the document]
RoyalPineapple: Another good question worth testing. It turns out that it is not ignored at all. 
Voiceover still announces "more content available", and when accessing the content through the rotor there is an "empty slot" where voiceover remains silent.

NickEntin:
The failing CI job should be fixed now with #209. If you rebase on latest main, that build should start passing.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilitySnapshotView.swift
```diff
@@ -447,11 +483,18 @@ private extension AccessibilitySnapshotView {
...
                 )
             }
-
+            
+            static func moreContentAvailableText(for locale: String?) -> String {
+                return "More Content Available".localized(
+                    key: "custom_content.description",
+                    comment: "Description for an accessibility element indicating that it has additional custom content available",
+                    locale: locale
+                )
```
NickEntin, line 492: This needs to be added to the localized strings files (why the Validate Localized Strings job is failing). There are schemes for the demo app to launch in German and Russian so you can pull the translations from what VoiceOver says.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilitySnapshotView.swift
```diff
@@ -594,6 +637,148 @@ private extension AccessibilitySnapshotView {
 
 }
 
+
+// MARK: -
+
+private final class CustomContentView: UIView {
```
meherkasam, line 643: @NickEntin Not an item for this PR, but this file is starting to get fairly large. Would it be worthwhile refactoring the subclasses into their own files?

meherkasam (approved):

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilitySnapshotView.swift
NickEntin: Yeah, it's probably time 👍

PR #177, opened 2024-01-04
Improve hit test performance by optimizing Y axis

NickEntin:

PR #119, opened 2023-03-12
Add utilities for highlighting hit test regions

NickEntin:
This introduces new snapshotting methods for snapshotting a view with its hit targets highlighted.

PR #178, opened 2024-01-04
Improve hit test performance by optimizing X axis

NickEntin:


Example/AccessibilitySnapshot.xcodeproj/xcshareddata/xcbaselines/607FACE41AFB9204008FA782.xcbaseline/EDA548A6-E453-4CBA-9366-15D413356833.plist
```diff
@@ -11,7 +11,7 @@
 				<key>com.apple.XCTPerformanceMetric_WallClockTime</key>
 				<dict>
 					<key>baselineAverage</key>
-					<real>17.200000</real>
+					<real>5.930000</real>
```
NickEntin, line 14: Setting both the permissible missed width and height to `1` gives another 65% improvement over #177, leading to a total ~88% improvement over #119.

NickEntin: Setting both to `2` brings it down to 1.961 sec.

Setting both to `4` brings it down to 1.074 sec.

dnagler (approved):
This is great, thanks!

Sources/AccessibilitySnapshot/Core/Swift/Classes/HitTargetSnapshotView.swift
```diff
@@ -111,32 +123,55 @@ public enum HitTargetSnapshotUtility {
...
+                            let stepHitView = view.hitTest(CGPoint(x: stepX + touchOffset, y: y), with: nil)
+                            updateForHit(stepHitView, at: stepX)
+                        }
                     }
                 }
 
                 // Finish the scan line if necessary.
-                if let lastHit = lastHit, let lastHitView = lastHit.1 {
-                    scanLine.append(((lastHit.0...bounds.maxX), lastHitView))
+                if lastHit.0 != bounds.maxX {
```
NickEntin, line 173: Oh wait, this has a similar problem to the Y axis. This assumes the `maxPermissibleMissedRegionWidth` will get you very close to the edge of the `bounds`, but if the width isn't nicely divisible by the step size this might not be true. I think we need another stride here to finish off the pass.

NickEntin: Vertical scroll indicator might be a good example of where this could fail right now

PR #182, opened 2024-01-12
Fix issue where SwiftUI toggles do not describe their accessibility v…

### LINKED ISSUE #181: Description for SwiftUI toggles missing accessibility value
I'd expect the description to include ", off" at the end to describe the toggle's description or ", on" if the toggle were on to match the description from VoiceOver.

Output from accessibility snapshot testing tool with the VoiceOver description for a SwiftUI toggle. The description generated does not include the value of "off"

DavidBrunow:
Fixed #181.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,4 +91,4 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
       - name: Build
-        run: carthage build --no-skip-current --platform iOS --use-xcframeworks
+        run: carthage build --no-skip-current --platform iOS --use-xcframeworks --verbose


Sources/AccessibilitySnapshot/Core/Swift/Classes/UIAccessibility+SnapshotAdditions.swift
```diff
@@ -113,7 +113,14 @@ extension NSObject {
                 } else {
                     traitSpecifiers.append(strings.switchButtonOffTraitName)
                 }
-
+            } else if let accessibilityValue = self.accessibilityValue {
```
DavidBrunow, line 116: I need to test more to ensure that this condition doesn't break the handling for the accessibility traits containing `.button` below.

DavidBrunow:
Not needed, this was fixed in November with #170.

Issue #51, opened 2020-12-10
Snapshotting with invert colors doesn't respect snapshot identifier


NickEntin:
Specifically the Swift version of the snapshot method in the `iOSSnapshotTestCase` subspec. The Objective-C version of this behaves correctly.

To reproduce:
```swift
func testView() {
    let view = UIView(frame: .init(x: 0, y: 0, width: 100, height: 100))
    SnapshotVerifyWithInvertedColors(view, identifier: "someIdentifier")
}
```
When run in record mode, this should produce a reference image named `testView_someIdentifier_` followed by the device properties. Instead it produces a reference image named `testView_` (omitting the identifier).

NickEntin:
This should be fairly straightforward to address (the actual fix is a one-liner).

I think it's worth updating the snapshot test a bit at the same time. Specifically, we should add an identifier to the test to verify this fix, and it's also probably worth adding some verification around the state restoration, e.g.:

```swift
    func testInvertColors() {
        let viewController = InvertColorsViewController()
        viewController.view.frame = UIScreen.main.bounds

        FBSnapshotVerifyView(viewController.view, identifier: "disabled")

        SnapshotVerifyWithInvertedColors(viewController.view)

        // Run the plain snapshot a second time to ensure that the view was restored to its original state.
        FBSnapshotVerifyView(viewController.view, identifier: "disabled")
    }
```

Due to a view life cycle issue with the snapshot test framework, you'll need to modify where the status get checked in the sample view controller in order to get the initial check to pass. The easiest way to accomplish this is to move the code from `viewWillAppear(_:)` to the end of `viewDidLoad()`.

PR #183, opened 2024-01-16
Include the identifier in the call to FBSnapshotVerifyView for SnapshotVerifyWithInvertedColors

kruegermj:
Resolves https://github.com/cashapp/AccessibilitySnapshot/issues/51


Example/AccessibilitySnapshot/InvertColorsViewController.swift
```diff
@@ -33,8 +33,8 @@ final class InvertColorsViewController: AccessibilityViewController {
         view = View()
     }
 
-    override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
+    override func viewDidLoad() {
+        super.viewDidLoad()
```
NickEntin, line 37: Why the change to `viewDidLoad()`? This should mirror the notification center observer removal in `viewWillDisappear(_:)` below.

kruegermj: I misunderstood the comment below in the [issue](https://github.com/cashapp/AccessibilitySnapshot/issues/51). I think we just need to update the status label in viewDidLoad. I made that change. 

> Due to a view life cycle issue with the snapshot test framework, you'll need to modify where the status get checked in the sample view controller in order to get the initial check to pass. The easiest way to accomplish this is to move the code from viewWillAppear(_:) to the end of viewDidLoad().

NickEntin (approved):

PR #191, opened 2024-01-20
Add Bazel build support for swift-snapshot-testing

luispadron:
Refactors the Bazel support to better match the Package.swift structure:

- The module `AccessibilitySnapshot_SnapshotTesting` uses `swift-snapshot-testing` and the `AccessibilitySnapshot_iOSSnapshotTestCase` modules use `iOSSnapshotTestCase`.
- Adds custom BUILD file for `swift-snapshot-testing` to allow it to build with Bazel. I opted for this vs. registry because it's a simple BUILD files.
- Adds `includes` so downstream targets can include the AccessibilitySnapshot headers via `#import <>`.


NickEntin:
> The module AccessibilitySnapshot uses swift-snapshot-testing and the FBSnapshotTestCase_Accessibility* modules use iOSSnapshotTestCase.

I think I'd rather be explicit about this (`AccessibilitySnapshot_iOSSnapshotTestCase` and `AccessibilitySnapshot_SnapshotTesting`) than assume the framework uses `SnapshotTesting`. The only reason the Package.swift defaults that way is iOSSnapshotTestCase didn't have SPM support when it was written. But seeing how things are going, I don't want to assume we'll stick with any particular snapshot engine for the long term. Eventually we may pull the snapshot comparison logic internally and only provide the `SnapshotTesting` integration as a convenience to use their method names.

luispadron:
Ah that's fair, I was copying the Package setup but we can make this explicit on the impl used.

luispadron:
@NickEntin I updated this to move the `BUILD.bazel` into the individual directories so its more explicit. 

Bazel users can depend on either `//Sources/Accessibility/SnapshotTesting/...` targets or ``//Sources/Accessibility/iOSSnapshotTestCase/...` targets. They would import it using the same  module name `import AccessibilitySnapshot` (let me know if we instead prefer this import name to also be explicit, I assumed users wouldn't use both implementations in the same target)

Sources/AccessibilitySnapshot/iOSSnapshotTestCase/BUILD.bazel
```diff
@@ -0,0 +1,74 @@
...
+    hdrs = glob(
+        ["ObjC/include/**/*.h"],
+        allow_empty = False,
+    ),
+    defines = ["BAZEL_PACKAGE"],
+    includes = ["ObjC/include"],
+    module_name = "AccessibilitySnapshot_ObjC",
+    tags = ["manual"],
+    visibility = ["//visibility:public"],
+    deps = [],
```
NickEntin, line 59: Likewise here, doesn't this depend on the Swift lib?

Sources/AccessibilitySnapshot/Core/BUILD.bazel
```diff
@@ -0,0 +1,73 @@
...
+    ),
+    data = glob(
+        ["Swift/Assets/**/*"],
+        allow_empty = True,
+    ),
+    defines = ["BAZEL_PACKAGE"],
+    module_name = "AccessibilitySnapshotCore",
+    tags = ["manual"],
+    visibility = ["//visibility:public"],
+    deps = [],
```
NickEntin, line 26: Doesn't this need to depend on `"AccessibilitySnapshotCore_ObjC.lib"`?

Sources/AccessibilitySnapshot/iOSSnapshotTestCase/BUILD.bazel
luispadron: They build without the dep and I can't see either depending on each other in the source

NickEntin: The macros in `FBSnapshotTestCase_Accessibility.h` (and the imprecise equivalents) calls the methods defined in the Swift code.

Sources/AccessibilitySnapshot/Core/BUILD.bazel
NickEntin: Even though there's no direct code usage, the Swift lib won't work correctly unless `+[ASAccessibilityEnabler load]` is run.

Sources/AccessibilitySnapshot/iOSSnapshotTestCase/BUILD.bazel
luispadron: Ah gotcha, I've fixed the deps. Once the examples use Bazel too it should be caught during tests

Sources/AccessibilitySnapshot/Core/BUILD.bazel
luispadron: Fixed this one too

NickEntin (approved):

NickEntin (approved):

PR #196, opened 2024-01-24
Support block-based accessibility properties

### LINKED ISSUE #192: Framework does not reflect `accessibility{Property}Block` values set in iOS 17
In iOS 17, the newly introduced `accessibility{Property}Block` properties will override the values from the corresponding `accessibility{Property}`.

NickEntin:
These block-based variants of the UIAccessibility properties were introduced in iOS 17. Resolves #192.

### DIFF
--- a/Sources/AccessibilitySnapshot/Core/Swift/Classes/UIAccessibility+EffectiveProperties.swift
+++ b/Sources/AccessibilitySnapshot/Core/Swift/Classes/UIAccessibility+EffectiveProperties.swift
@@ -1,0 +1,227 @@
+//
+//  Copyright 2024 Block Inc.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


Sources/AccessibilitySnapshot/Core/Swift/Classes/UIAccessibility+EffectiveProperties.swift
```diff
@@ -0,0 +1,227 @@
...
+
+extension NSObject {
+
+    var effectiveIsAccessibilityElement: Bool {
+        #if swift(>=5.9)
+        if #available(iOS 17.0, *) {
+            return isAccessibilityElementBlock?() ?? isAccessibilityElement
+        } else {
+            return isAccessibilityElement
+        }
```
RunningBack33, line 29: you should remove 'else' block, you have 'return isAccessibilityElement' in the end, this is duplicate. 
And in other variables same

PR #201, opened 2024-01-26
Add CI job to validate localized strings files

NickEntin:
This ensures we don't forget to update a strings file when adding new descriptions.


kruegermj (approved):
🚀 

Does this confirm that for every localized string defined in a swift file there is a matching key in a strings file?

NickEntin:
> Does this confirm that for every localized string defined in a swift file there is a matching key in a strings file?

Yep, exactly! And also that there aren't duplicate keys.

RoyalPineapple (approved):

PR #203, opened 2024-01-27
Add German and Russian locales to demo app

NickEntin:
Listing these as supported locales and adding the associated schemes makes it easier to run the demo app to get the translated strings from VoiceOver.


meherkasam (approved):
Are you planning to add support for more locales over time? Could you use support for that?

NickEntin:
@meherkasam I'm not proactively adding more languages right now, but happy to add more languages as they're requested by developers. Each one adds some overhead work to making updates, so I don't want to add languages that consumers don't actually need.

PR #207, opened 2024-02-16
Update actions in CI to version using Node 20

NickEntin:
The current versions we're using are on Node 16, which is now deprecated. This updates us to the latest version of each action, which uses Node 20.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: '3.2.2'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -29,10 +29,10 @@ jobs:
       - name: Build and Test
         run: Scripts/build.swift xcode ${{ matrix.platform }} `which xcpretty`
       - name: Upload Results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: failure()
         with:
-          name: Test Results
+          name: Test Results (${{ matrix.platform }})
```
NickEntin, line 35: `upload-artifact` no longer supports uploading multiple artifacts with the same name

NickEntin:
Opened #210 to verify the new upload-artifact naming works as expected

kruegermj (approved):

PR #213, opened 2024-03-11
Sketch of VoiceOver SwiftUI previews

DavidBrunow:
This pull request is a sketch of what VoiceOver SwiftUI previews could look like.

Here are some decisions I made in this implementation:

* Created a modifier that could be added on top of an existing preview so multiple previews were not needed. This means that a button must be pressed, or a binding must be set to true, to see the VoiceOver preview
* Used the existing `AccessibilitySnapshotView`, including the snapshot of the content, rather than rendering the bounding boxes on top of the existing view
* Placed the legend below the view that has been snapshotted 

Here is a screen recording of the preview in action:

https://github.com/cashapp/AccessibilitySnapshot/assets/1452879/67ecc3f9-8264-41f5-8bbc-bf6e0b450ed6

### DIFF
--- a/AccessibilitySnapshotCore.podspec
+++ b/AccessibilitySnapshotCore.podspec
@@ -1,0 +1,21 @@
+Pod::Spec.new do |s|
+  s.name             = 'AccessibilitySnapshotCore'
+  s.version          = '0.6.0'
+  s.summary          = 'Easy regression testing for iOS accessibility'
+
+  s.homepage         = 'https://github.com/CashApp/AccessibilitySnapshot'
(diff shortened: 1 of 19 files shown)


AccessibilitySnapshotCore.podspec
```diff
@@ -0,0 +1,21 @@
+Pod::Spec.new do |s|
```
DavidBrunow, line 1: I had to create a new podspec to be able to import `AccessibilitySnapshotCore` into the code files with the previews.

PR #214, opened 2024-03-13
Remove h3’s from detail summary because GitHub doesn’t support them

ZevEisenberg:
The formatting was all weird, and these were probably too big anyway. Old at the top, new at the bottom:

image


RoyalPineapple (approved):
Much nicer, thanks.

PR #228, opened 2024-11-20
Add accessibility identifier in AccessibilityMarker

priip:
Adds a new `identifier` property to the `AccessibilityMarker` object which stores the [`accessibilityIdentifier`](https://developer.apple.com/documentation/uikit/uiaccessibilityidentification/1623132-accessibilityidentifier) for an element. This unique identifier can be used for identifying elements in UITests and is not visible to users.

### DIFF
--- a/Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
+++ b/Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
@@ -15,6 +15,7 @@
 //
 
 import Accessibility
+import SwiftUI
 import UIKit
 
 public struct AccessibilityMarker {
@@ -36,6 +37,10 @@ public struct AccessibilityMarker {
     /// The description of the accessibility element that will be read by VoiceOver when the element is brought into
     /// focus.
     public var description: String
+
+    /// A unique identifier for the element, primarily used in UI tests for locating and interacting with elements.
+    /// This identifier is not visible to users.
+    public var identifier: String?
 
     /// A hint that will be read by VoiceOver if focus remains on the element after the `description` is read.
     public var hint: String?
@@ -209,6 +214,7 @@ public final class AccessibilityHierarchyParser {
 
             return AccessibilityMarker(
                 description: description,
+                identifier: element.object.identifier,
                 hint: hint,
                 userInputLabels: userInputLabels,
                 shape: accessibilityShape(for: element.object, in: root),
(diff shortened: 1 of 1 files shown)


Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -741,6 +747,31 @@ fileprivate extension NSObject {
         }
         return []
     }
+
+    var identifier: String? {
+        if let accessibilityIdentifier = value(forKey: "accessibilityIdentifier") as? String {
+            return accessibilityIdentifier
+        }
+        return nil
+    }
```
priip, line 756: There is no "clean" way to access the `accessibilityIdentifier` for an `NSObject`, since it's only available for objects that conform to `UIAccessibilityIdentification` ([doc](https://developer.apple.com/documentation/uikit/uiaccessibilityidentification)) I tried using the `accessibilityIdentifierBlock`, but that yielded no results, so I've opted to pull the value using key values. I'm open to suggestions though if ya'll have any!

priip: I tried this, but it was not triggering any of the UIKIt views.

meherkasam: Ugh, okay yes I vaguely remember now having a discussion with @RoyalPineapple about something very similar. I forget the context, but perhaps he can recall what he ended up doing here.

RoyalPineapple: I think the solution here of testing for known conformance before using the key/value as a fallback works well enough. This bug is frustrating every time it comes up.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -714,7 +720,7 @@ extension UIView {
 fileprivate extension NSObject {
     var customContent: [(label: String, value: String, isImportant:Bool)] {
         // Github runs tests on specific iOS versions against specific versions of Xcode in CI.
-        // Forward deployment on old versions of Xcode require a compile time check which require diferentiation by swift version rather than iOS SDK.
+        // Forward deployment on old versions of Xcode require a compile time check which require differentiation by swift version rather than iOS SDK.
```
RoyalPineapple, line 723: thanks 👍🏻

RoyalPineapple (approved):

OfficialFedar (commented):
Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift

Issue #229, opened 2024-12-03
Support for packages without host application.


michael-zimmermann:
It is not possible to run accessibility snapshot testing (using point-free ios-snapshotting) when creating swift packages. 
The snapshots crash with 
```
AccessibilitySnapshot/SnapshotTesting+Accessibility.swift:57: Fatal error: Accessibility snapshot tests cannot be run in a test target without a host application
```

It doesn't appear that the host application is necessary, at least with the point-free framework. Here is a snapshot where the default guard causing the fatal error is overridden.

<img src="https://github.com/user-attachments/assets/e46d5b0f-771b-4eb9-be29-1c7db6aaf7b1" width="300"/>

I propose, to be able to pass a variable to the snapshotting-strategy to avoid checking if a host application is available. The default value `true` of this new parameter would keep the old strategy working the way it did and cause no backwards compatibility problems. By orverriding this to `false`, accessibility snapshots could also be done for swift packages.

```
        assertSnapshot(
            of: view,
            as: .accessibilityImage(
                shouldRunInHostApplication: false
            )
        )
```

szotp:
The behavior with host application is strange.

parseAccessibilityElements works just fine with UIHostingController, does not matter if hosting app is present or not

But with regular UIView, it returns nothing. Even with code as simple as
```swift
let label = UILabel()
label.text = "Hello"
XCTAssertEqual(label.accessibilityLabel, "Hello") // fails without host app
```

PR #232, opened 2025-01-13
Flag to avoid prerequisite of tests having to run in a host application.

michael-zimmermann:
See issue https://github.com/cashapp/AccessibilitySnapshot/issues/229


chwo:
@RoyalPineapple Pinging you since you published the last releases.

I am experiencing the same issue that the accessibilityImage assertion crashes when run inside SPM. I can confirm that this is solved by this PR so I would love to see that fix merged. Is there anything blocking this?

albsala-backmarket:
Hi there! Is there any progress here? We'd appreciate if this is fixed. Thank you!
cc @RoyalPineapple @NickEntin

albsala-backmarket:
Hey @RoyalPineapple thank you for the approval ❤️! 
Not sure if we need something else to merge this PR and close the issue?

NickEntin:
Apologies for the slow response here. This change feels like it introduces a pretty big footgun into the library. As @szotp's example in the linked issue illustrates, the **behavior when running outside a host application is unpredictable and can give some very deceptive test results**. While it's definitely an unfortunate requirement, I'd highly recommend reverting this change, since setting this parameter to false essentially means you can't trust the correctness of the resulting snapshots.

PR #237, opened 2025-02-27
fixed a bug where an element's accessibility value could be omitted

RoyalPineapple:


Sources/AccessibilitySnapshot/Core/Swift/Classes/UIAccessibility+SnapshotAdditions.swift
```diff
@@ -126,8 +126,9 @@ extension NSObject {
             case "2":
                 traitSpecifiers.append(strings.switchButtonMixedStateName)
             default:
-                // When the switch button trait is set, unknown accessibility values are omitted from the description.
```
RoyalPineapple, line 129: This may have previously been the case but since the trait was made public elements with unknown labels are announced the same as these magic values.

kyleve (approved):

meherkasam (approved):

Sources/AccessibilitySnapshot/Core/Swift/Classes/UIAccessibility+SnapshotAdditions.swift
NickEntin: FYI in the past we've tried to match behavior to the OS version, so e.g. figuring out when this behavior changed and diverging logic on the version here.

Don't have strong feelings about keeping that policy vs. saying we should keep it simple and always match current OS; only strong stance is we should decide on a policy and stick with it.

RoyalPineapple: Yea thats reasonable policy I'll put something together.

Issue #221, opened 2024-08-20
Sort ordering does not match VoiceOver order


g-mark:
A difference in accessibility snapshots was noticed when one of two side by side items changed from providing an outdented `acessibilityPath` to just using the default `accessibilityFrame`.

Further investigation found that the `AccessibilityHierarchyParser` has two potential issues with how elements are ordered in comparison with how VoiceOver works.
1. The ordering of elements is based on either the `accessibilityPath` (if not `nil`) or the `accessibilityFrame`. VoiceOver seems to always use the `accessibilityFrame`.
2. The vertical y comparison of elements uses a strict != comparison, whereas VoiceOver has about a ± 7 pt "roughly equal to" comparison.

This view controller shows some UILabels where one provides a custom `accessibilityFrame`, and another an `accessibilityPath` - each insets the view's frame by a different amount. Below is the VoiceOver focus order as seen by the Accessibility Inspector.
| Header | "Lorem ipsum" label | "frame" label | "path" label |
|---|---|---|---|
| <img width="420" alt="image" src="https://github.com/user-attachments/assets/c36ed22c-071e-4985-856b-cc7a8edad58c"> | <img width="420" alt="image" src="https://github.com/user-attachments/assets/0fe3046c-9865-4ae8-a982-d847e23028bf"> | <img width="421" alt="image" src="https://github.com/user-attachments/assets/25290d8a-8401-4e82-81ba-8b55d14319b8"> | <img width="420" alt="image" src="https://github.com/user-attachments/assets/66223ffc-e573-469b-8458-ca9a097986d1"> |

In the snapshot image, however the order is: Header, "path" label, "frame" label, then "Lorem ipsum" label.
![image](https://github.com/user-attachments/assets/7b7ec2b4-fff4-4eed-99f5-546111e63fd0)

Sample project: [AccessibilityElementOrder.zip](https://github.com/user-attachments/files/16678868/AccessibilityElementOrder.zip)

RoyalPineapple:
Thanks for the great sample project.
I verified that the accessibility path is completely ignored for sorting purposes on devices going back to iOS 15.

PR #242, opened 2025-03-13
Adjusting sort order to more closely match VoiceOver

RoyalPineapple:
This PR resolves Issue https://github.com/cashapp/AccessibilitySnapshot/issues/221 which points out two subtle details in VoiceOver's positional element sorting.

- VoiceOver element sorting always uses accessibility frames, even when a path is present. 
- VoiceOver treats accessibility frames with an origin.y delta of less than 8 to be in line with one another and eligible for sorting by leading -> trailing. 

I've confirmed this behavior on device as far back as iOS15. This does not appear to be a change in VoiceOver behavior from any previous versions rather a detail that has been overlooked in our implementation. So we won't need to fork the behavior by version. 

This also falls under the 'Correcting Accessibility Element Descriptions' versioning guidance making this change a minor version bump.


kyleve (approved):

Example/SnapshotTests/ReferenceImages/_64/SnapshotTests.ElementFrameComparisonTests/testFrames_16_4_393x852@3x.png
meherkasam, line 1: If `path down` is 8 pt below `path unchanged`, shouldn't `fifth` come before `fourth`?

RoyalPineapple: path is completely ignored for these calculations.

PR #244, opened 2025-05-27
Updated Sort Order

RoyalPineapple:
Based on discussions with @meherkasam  this PR updates the sort order to take the userinterfaceidiom into account.

### DIFF
--- a/Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
+++ b/Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
@@ -271,11 +271,13 @@ public final class AccessibilityHierarchyParser {
     /// this should typically be `false`.
     /// - parameter root: The root view to which the nodes' shapes are relative.
     /// - parameter userInterfaceLayoutDirection: The device's current user interface layout direction.
+    /// - parameter userInterfaceIdiom: the device's interface idiom, used to calculate the sort order
     private func sortedElements(
         for nodes: [AccessibilityNode],
         explicitlyOrdered: Bool,
         in root: UIView,
-        userInterfaceLayoutDirection: UIUserInterfaceLayoutDirection
+        userInterfaceLayoutDirection: UIUserInterfaceLayoutDirection,
+        userInterfaceIdiom: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom
     ) -> [Element] {
         // VoiceOver flick navigation iterates through elements in a horizontal, then vertical order. The horizontal
         // ordering matches the application's user interface layout direction. The vertical ordering is always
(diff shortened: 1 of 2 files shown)


Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -300,9 +315,9 @@ public final class AccessibilityHierarchyParser {
             fatalError("Unknown user interface layout direction: \(userInterfaceLayoutDirection)")
         }
         
-        // 8 seems to be the magic number for VoiceOver to consider it
-        // to be vertically "above" other views.
-        let minimumVerticalSeparation = 8.0
+        // Derived via experimentation, these magic numbers are the cutoff for VoiceOver to consider
+        // an element to be vertically "above" other views.
+        let minimumVerticalSeparation = userInterfaceIdiom == .phone ? 8.0 : 12.0
```
meherkasam-square, line 320:
```suggestion
        let minimumVerticalSeparation = userInterfaceIdiom == .phone ? 8.0 : 13.0
```

meherkasam-square (approved):

meherkasam (approved):

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -171,17 +180,20 @@ public final class AccessibilityHierarchyParser {
     /// In most cases, this should use the default value, `UIApplication.shared`.
     public func parseAccessibilityElements(
         in root: UIView,
-        userInterfaceLayoutDirectionProvider: UserInterfaceLayoutDirectionProviding = UIApplication.shared
+        userInterfaceLayoutDirectionProvider: UserInterfaceLayoutDirectionProviding = UIApplication.shared,
+        userInterfaceIdiomProvider: UserInterfaceIdiomProviding = UIDevice.current
```
kyleve, line 184: Noting that in the Market snapshot tests we will want to inject a different `UserInterfaceIdiomProviding`, since they're kinda magic: They can run on _any_ 3x iOS device and do swizzling and such to simulate all different device sizes, safe areas, etc, that we need.

Sources/AccessibilitySnapshot/Core/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -300,9 +315,9 @@ public final class AccessibilityHierarchyParser {
             fatalError("Unknown user interface layout direction: \(userInterfaceLayoutDirection)")
         }
         
-        // 8 seems to be the magic number for VoiceOver to consider it
-        // to be vertically "above" other views.
-        let minimumVerticalSeparation = 8.0
+        // Derived via experimentation, these magic numbers are the cutoff for VoiceOver to consider
+        // an element to be vertically "above" other views.
+        let minimumVerticalSeparation = userInterfaceIdiom == .phone ? 8.0 : 13.0
```
kyleve, line 320: Not for this PR, but I am wondering if Voiceover is also a bit more clever when seeing a hierarchy like this:

```
┌──────────────────────────────────┐
│                                  │
│             ┌───────┐            │
│  ┌───────┐  │2      │ ┌───────┐  │
│  │1      │  │       │ │3      │  │
│  └───────┘  │       │ │       │  │
│             └───────┘ │       │  │
│                       └───────┘  │
│                                  │
│                                  │
│  ┌───────┐                       │
│  │4      │  ┌───────┐            │
│  └───────┘  │5      │ ┌───────┐  │
│             │       │ │6      │  │
│             │       │ └───────┘  │
│             └───────┘            │
│                                  │
└──────────────────────────────────┘
```

Eg does it try to also group content into "rows" (eg, 1, 2, 3 would be in one group, 4, 5, 6) in another, and then orders the items in the groups LTR/RTL based on layout direction, rather than purely by vertical separation? That would make a bit more sense to me.

RoyalPineapple:
> Will this fix #129 and #168 like #169 does?

Probably not unfortunately, this is an unrelated fix.

PR #247, opened 2025-06-25
Pin GitHub actions

meherkasam:
I ran the `ratchet` command on `ci.yml` to pin the GitHub actions.


soroushsq (approved):
I think it's fine to pin all the versions, not just the ones from the `ruby` org! Let's wait for the morning to let @RoyalPineapple chime in, but otherwise this looks fine

RoyalPineapple (approved):

PR #249, opened 2025-07-08
Publish version 0.10.0

meherkasam:


soroushsq (approved):
This is good with me, but let's make sure to push this version to the central cocoapods repo as well. We also need to make a tag for SPM, right?

meherkasam:
> This is good with me, but let's make sure to push this version to the central cocoapods repo as well. We also need to make a tag for SPM, right?

I have a draft release ready for SPM. As for CocoaPods, I will get on that.

PR #250, opened 2025-07-16
Extract Parser module out of Core

soroushsq:


meherkasam (approved):

Sources/AccessibilitySnapshot/Core/ViewRenderingMode.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  ViewRenderingMode.swift
+//  AccessibilitySnapshot
+//
+//  Created by Soroush Khanlou on 7/1/25.
+//
+
+/// TODO: Docs
```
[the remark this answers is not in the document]
soroushsq: I don't even remember adding this TODO but I'm glad I did

Sources/AccessibilitySnapshot/Core/DefaultColors.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  DefaultColors.swift
+//  AccessibilitySnapshot
+//
+//  Created by Soroush Khanlou on 7/1/25.
+//
+
+import UIKit
+
+public enum MarkerColors {
```
kyleve, line 10: I love the Marker design system

Sources/AccessibilitySnapshot/Core/Localizations.swift
kyleve, line 1: Do we need to re-extract or move any strings files?

kyleve: Nvm missed those below!!

kyleve (approved):

Package.swift
```diff
@@ -27,6 +27,10 @@ let package = Package(
             name: "AccessibilitySnapshotCore",
             targets: ["AccessibilitySnapshotCore"]
         ),
+        .library(
+            name: "AccessibilitySnapshotParser",
```
RoyalPineapple, line 31: How much work is it going to be if we want to rename these things in the future?

Sources/AccessibilitySnapshot/Core/DefaultColors.swift
```diff
@@ -0,0 +1,12 @@
...
+    public static let defaultColors: [UIColor] = [ .cyan, .magenta, .green, .blue, .yellow, .purple, .orange ]
```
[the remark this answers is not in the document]
soroushsq: We can, but this will break existing screenshots, which I don't think we want to do yet.

PR #263, opened 2025-09-18
[DNM] [DNR] Accessibility order markers

RoyalPineapple:
This change adds overlay markers that explicitly expose the accessibility sort order providing greater accessibility for colorblind users.
This can be also be useful to quickly validate ordering without needing to continually reference the legend view.


Example/SnapshotTests/ReferenceImages/_64/SnapshotTests.ElementOrderTests/testGrid_16_4_393x852@3x.png
RoyalPineapple, line 1: These markers may be too subtle, its a delicate balance. I don't want to overpower the element under test.

PR #275, opened 2025-11-19
Add iOS 17.5 and 18.5 test support, remove iOS 16.4

RoyalPineapple:
Add support for testing on iOS 18.5 simulator with snapshot reference images.

This PR:
- Adds iOS 18.5 snapshot reference images
- Updates iOS 17.5 snapshot reference images  
- Removes iOS 16.4 snapshot reference images (no longer supported)

iOS 26 support will be added in a separate PR.


.github/workflows/ci.yml
```diff
@@ -8,11 +8,17 @@ on:
...
     strategy:
       matrix:
-        platform: ['iOS_17', 'iOS_16']
+        include:
+          - platform: iOS_17
+            runner: macOS-14
+          - platform: iOS_18
+            runner: macOS-15
+          - platform: iOS_26
+            runner: macOS-15
```
soroushsq: I guess we can't run 26 on 26 because github hasn't released macos 26 runners yet?

PR #278, opened 2025-12-02
Add container visualization for accessibility snapshots

RoyalPineapple:
Adds container visualization to accessibility snapshot legends without changing element overlay behavior.

## Summary
- New `showContainers` configuration flag turns on a hierarchical legend that groups elements by their accessibility container with dashed borders and per-container badges
- Snapshot image and element overlays are baked into a single flat `UIImage` before the legend enters layout, so overlay positions can never be shifted by legend placement
- Parser/render order swap fixes a pre-existing overlay misalignment for nested `UIHostingController` content (e.g. `SemanticGroupWrapper`)
- Demo coverage for every container shape the parser produces (semanticGroup × 4 info variants, list, landmark, dataTable, tabBar) plus a SwiftUI `Table` reveal
- iOS 26.4 reference images recorded for the new tests

### Snapshot + overlay baking
`PreParsedAccessibilitySnapshotView.init` composites the snapshot image and all element overlays into a single flat `UIImage` up front. The rendered `snapshotWithOverlays` body is just `Image(uiImage:)` — no `ZStack` of overlays on top of the image. This decouples overlay alignment from the surrounding SwiftUI layout, so the legend (right, below, multi-column, future flow layouts) can never shift overlay coordinates.

Baking uses `drawHierarchy(afterScreenUpdates: true)` inside a temporary `UIWindow` with `safeAreaRegions = []` so SwiftUI content renders correctly without safe-area offsets.

### Overlay alignment fix
`drawHierarchy(afterScreenUpdates: true)` mutates layout mid-render for views that defer layout until the render cycle — notably nested `UIHostingController` instances inside `UIViewRepresentable` wrappers (e.g. `SemanticGroupWrapper`). When parsing happened after rendering, the parser read `accessibilityFrame` values from a post-render layout that no longer matched what the image captured, shifting the first group's overlay.

Fix: swap the parse/render order in `AccessibilitySnapshotBaseView.parseAccessibility()` so marker frames and the rendered image both come from the same pre-render layout state.

### Container legend
When `showContainers` is enabled, the legend renders a hierarchical view that groups elements by their accessibility containers using dashed borders and container badges. Container entries expand to the full legend column width so multi-word badges like "Data Table (3 × 4)" stay on one line. The snapshot area is identical regardless of `showContainers` — only the legend content changes.

Supported container types in the badge label:
- `semanticGroup(label, value, identifier)` → "label: value", or identifier, or "Semantic Group"
- `list` → "List"
- `landmark` → "Landmark"
- `dataTable(rows, cols)` → "Data Table (r × c)"
- `tabBar` → "Tab Bar"

New types:
- `HierarchyColorAssignment` — assigns color indices to hierarchy nodes (elements use traversal-order position matching overlay indices; containers use a separate counter)
- `ContainerLegendEntryView` / `ContainerBadge` — dashed-border legend entry with layered-icon badge
- `HierarchyLegendView` — recursive hierarchical legend renderer

### Demos and tests
Each container type now has a focused demo + test under `AccessibilitySnapshotPreviewsDemo`:

- `ContainerDemo` — four `semanticGroup` info variants (label only, label + value, identifier only, all three)
- `ListContainerDemo` — `accessibilityContainerType = .list`
- `LandmarkContainerDemo` — `accessibilityContainerType = .landmark`
- `TabBarContainerDemo` — `accessibilityTraits = .tabBar` on a non-`UITabBar` view
- `DataTableContainerDemo` — `accessibilityContainerType = .dataTable` with real `UIAccessibilityContainerDataTableCell` elements reporting per-cell row/column ranges, so the legend shows "Data Table (3 × 4)" plus each cell's row/column context
- `SwiftUITableDemo` — exposes how SwiftUI's native `Table` collapses to a one-column List on iPhone; the parser still emits a `.dataTable` container but with `rowCount = 0, columnCount = 0` and no `UIAccessibilityContainerDataTableCell` elements

Shared `AccessibilityContainers.swift` provides:
- A generic `UIViewRepresentable` that wraps SwiftUI content in a `UIView` reporting a configured `UIAccessibilityContainerType` or the `.tabBar` trait
- An `AccessibilityDataTable` `UIViewRepresentable` that renders a 2D grid of labels and exposes each cell as a real `UIAccessibilityElement` conforming to `UIAccessibilityContainerDataTableCell`

### Other changes
- `FBSnapshotTestCase+SwiftUI`: sets `hostingController.safeAreaRegions = []` on iOS 16.4+ in the config-based overload, preventing `drawHierarchy` from baking status bar / safe-area offsets into the snapshot
- `AccessibilitySnapshotConfiguration`: new `showContainers: Bool = false` flag
- `ParsedAccessibilityData`: now also carries the full `hierarchy: [AccessibilityHierarchy]` tree (not just flattened markers) so the container legend can render

## Test plan
- [x] All existing snapshot tests pass
- [x] Six new container-demo tests produce overlays aligned 1:1 with snapshot content
- [x] Container legend correctly groups elements with dashed borders and full-width badges
- [x] DataTable demo reports per-cell row/column context (no NSNotFound noise)
- [x] iOS 26.4 reference images recorded for all new tests


RoyalPineapple (commented):
Still in progress

RoyalPineapple:
🤖 Superseded by upstream container support (#339, #343) and fork-side container visualization work.

PR #294, opened 2026-01-22
allowing dismissing of view controllers in the example app

soroushsq:
Something small that came up on a call with me and @RoyalPineapple today, you can't dismiss view controllers after they've been presented, you have to kill the app. This fixes that.


meherkasam-square (approved):

meherkasam-square:
Thanks for fixing this!

Issue #223, opened 2024-09-27
Support for Accessibility Containers


cnstoll:
Hi! I was wondering if ya'll have considered support for Accessibility Containers. The main reason I ask is that containers can have an accessibility label and custom actions that are still read even though the container itself isn't a focusable accessibility element. I feel like it would be useful to having that data in the snapshot images. I might be able to implement this change, but I wanted to know if ya'll have considered it or if there's any reason it wasn't implemented already. Thanks!

PR #325, opened 2026-03-26
Fix section header/footer ordering in SwiftUI List snapshots

RoyalPineapple:
## Summary

Section headers and footers in SwiftUI `List` with `Section` are rendered in the wrong order — all row cells appear first, then all headers/footers are grouped at the end. VoiceOver reads them in the correct interleaved order.

**Root cause:** The parser computes a group's sort position using the **union** of all children's frames. When `UICollectionView` (which backs SwiftUI `List` on iOS 16+) stores headers and cells in separate internal container views, the union-frame for each container spans the full list height, making their relative sort order arbitrary. Additionally, groups created from `accessibilityElements` were always marked `explicitlyOrdered: true`, preventing frame-based re-sorting even when the array only contained sub-groups (not direct elements).

**Fix (two changes in `AccessibilityHierarchyParser.swift`):**

1. **`accessibilitySortFrame` for `.group`**: Use the topmost child's frame instead of the union of all children. This matches the comment in `sortedElements` about VoiceOver positioning groups by their "first element."

2. **`explicitlyOrdered` for `accessibilityElements` groups**: Only mark as explicitly ordered when the group contains direct `.element` nodes. When it contains only sub-groups (as with `UICollectionView`'s internal containers), allow frame-based re-sorting to correctly interleave headers and cells.

## Breaking change

The sort-order fix changes the element ordering returned by `parseAccessibilityElements(in:)` for views that use accessibility containers (e.g. SwiftUI `List`, `Form`, `NavigationStack`). The new order is more correct — it matches VoiceOver's actual traversal — but **existing snapshot baselines for these views will need to be re-recorded**.

Views without accessibility containers are unaffected.

### `List` with section headers

| Before (broken) | After (fixed) |
|---|---|
| before-headers | after-headers |

**Before:** Apple, Banana, Cherry, Carrot, Peas, _then_ FRUITS Heading, VEGETABLES Heading
**After:** FRUITS Heading, Apple, Banana, Cherry, VEGETABLES Heading, Carrot, Peas

### `List` with headers and footers

| Before (broken) | After (fixed) |
|---|---|
| before-footers | after-footers |

**Before:** Checking, Savings, Electric, Internet, _then_ ACCOUNTS Heading, footer, BILLS Heading, footer
**After:** ACCOUNTS Heading, Checking, Savings, footer, BILLS Heading, Electric, Internet, footer

## How it works

```
UICollectionView (accessibilityElements = [containerA, containerB, containerC])
├── containerA (cells):     [Apple, Banana, Cherry, Carrot, Peas]     ← union frame spans full list
├── containerB (headers):   [FRUITS heading, VEGETABLES heading]       ← union frame spans full list
└── containerC (footers):   [footer1, footer2]                         ← union frame spans full list
```

**Before:** Each container's sort frame = union of children = full list height → containers sort arbitrarily → cells before headers.

**After:** Each container's sort frame = topmost child's frame → containerB sorts first (FRUITS heading is at top) → headers interleave correctly with cells.

## Related issues

- Related to #129 — Incorrect sort order for SwiftUI Form in iOS 16. The two-part fix in this PR was tested against `Form` and **does not fully resolve** the issue: headers still group at the end of the legend. Form has additional internal nesting that requires separate investigation.
- Related to #168 — Incorrect sort order when using NavigationStack. Local repro shows the toolbar button is missing from the accessibility hierarchy entirely — a separate discovery bug rather than a sort ordering bug.
- Related to #223 — Support for Accessibility Containers.

## Test plan

- [x] Added `SwiftUIListSectionTests` with two snapshot tests covering headers-only and headers+footers cases
- [x] Added `SwiftUIListWithSections` and `SwiftUIListWithHeadersAndFooters` demo views to the app for manual VoiceOver testing
- [x] Verified VoiceOver traversal order matches the fixed snapshot order on a real device
- [x] Unit tests in `AccessibilityHierarchyParserTests` verify the sort frame computation


johnnewman-square (approved):
This looks good! A preview demo with a List would be a neat addition, too.

Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -649,7 +645,14 @@ private extension AccessibilityHierarchyParser {
...
         case let .group(elements, _, _, _):
-            return elements.reduce(CGRect.null) { $0.union(accessibilitySortFrame(for: $1, in: root)) }
+            // Matches VoiceOver behavior: groups sort by their first-selected child
+            // (see comment in sortedElements).
+            return elements
+                .map { accessibilitySortFrame(for: $0, in: root, horizontalCompare: horizontalCompare) }
+                .min { f1, f2 in
+                    if f1.origin.y != f2.origin.y { return f1.origin.y < f2.origin.y }
+                    return horizontalCompare(f1.origin.x, f2.origin.x)
+                } ?? .null
```
robmaceachern, line 658: suggestion: Reuse the thresholded row-order logic here

This helper picks a group's representative child using raw `y` before `x`, but `sortedElements` only gives vertical precedence once the delta exceeds the 8pt/13pt threshold. In a near-threshold layout, this can choose a different "first" child than the parent sorter would and shift the whole group relative to siblings. Reusing the same thresholded comparator here, plus a regression test for that boundary, would keep the group sort frame aligned with the actual traversal logic.

---

Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -720,11 +723,19 @@ private extension NSObject {
...
+            // separate internal containers (e.g. UICollectionView headers vs cells).
+            let hasDirectElements = accessibilityHierarchyOfElements.contains {
+                if case .element = $0 { return true }
+                return false
+            }
             recursiveAccessibilityHierarchy.append(.group(
                 accessibilityHierarchyOfElements,
-                explicitlyOrdered: true,
+                explicitlyOrdered: hasDirectElements,
                 frameOverrideProvider: overridesElementFrame(with: contextProvider) ? self : nil,
```
robmaceachern: blocker: This changes the explicit ordering contract for accessibility containers

This now switches `accessibilityElements` containers from "preserve the container-provided order" to "re-sort by frame unless an immediate child is a direct element." That fixes the `UICollectionView` structural-artifact case, but it also changes behavior for custom containers that intentionally return subgroup views in semantic order rather than visual order. The surrounding comment in `sortedElements` still says container traversal follows the order specified by the container, so I think this needs a narrower opt-in for the broken UIKit case or a regression test proving intentional subgroup ordering is still preserved.

---

robmaceachern (approved):
Included some AI feedback for your consideration but lgtm

Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
RoyalPineapple: Good catch — fixed in 7d68ef41. `accessibilitySortFrame` now takes `minimumVerticalSeparation` and uses the same thresholded comparison as `sortedElements`. Added `testGroupSortFrameRespectsVerticalThreshold` to cover the boundary case (two groups within the 8pt phone threshold, horizontal tiebreaker).

Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
RoyalPineapple: You were right — reverted in 7d68ef41.

Tested on a real device (iPhone 15 Pro) using `UIAccessibility.elementFocusedNotification` to capture VoiceOver's actual traversal order across several scenarios:

**Finding: VoiceOver always preserves `accessibilityElements` array order**, regardless of whether children are direct elements or subgroups. When we set `accessibilityElements = [cellContainer, headerContainer]`, VoiceOver reads cells first, then headers — matching the array order, not the visual layout.

**The real UICollectionView fix works differently.** UICollectionView's internal container views (for cells, headers, footers) are discovered via the subview walk, not through `accessibilityElements`. Those groups already have `explicitlyOrdered: false`, so the sort-frame change (topmost child instead of union) is sufficient to correctly interleave them.

Changes:
- Reverted `explicitlyOrdered` back to `true` for all `accessibilityElements` groups
- Updated `testAccessibilityElementsPreservesOrderEvenWithOnlySubgroups` to assert the correct (array-preserving) order
- The sort-frame fix and threshold fix remain as the actual fix for the List section ordering bug

PR #326, opened 2026-03-29
Add filter, map, reduce tree operations for AccessibilityHierarchy

RoyalPineapple:
## Summary

Adds first-class functional tree operations to `AccessibilityHierarchy`:

- **`filter`** — prunes subtrees, preserving container structure for surviving nodes
- **`map`** — transforms every node bottom-up so closures see already-mapped children
- **`reduce`** — folds the tree into a single value via pre-order depth-first traversal

Plus `filteredHierarchy`/`mappedHierarchy`/`reducedHierarchy` array conveniences for operating across multiple roots.

## Motivation

The existing tree utilities (`forEach`, `sortIndex`, `flattenToElements`, `flattenToContainers`) each hand-roll their own recursion with `switch`/`case` boilerplate. This PR extracts the recurring pattern into three composable primitives, then rewrites the existing utilities in terms of them — eliminating manual recursion while preserving all public API signatures.

## What changed

| File | Change |
|------|--------|
| `AccessibilityHierarchy+TreeOperations.swift` | New file: `filter`, `map`, `reduce` + array conveniences |
| `AccessibilityHierarchy.swift` | `sortIndex`, `forEach`, `flattenToElements`, `flattenToContainers` rewritten using the new primitives |
| `AccessibilityHierarchyTreeOperationsTests.swift` | 40+ tests: per-operation coverage, edge cases, composition chains, and equivalence validation proving the rewrites match original behavior |
| `Package.swift` | New `AccessibilitySnapshotParserTests` test target |

## Test plan
- [x] All new tests pass on iOS Simulator
- [x] Full project build succeeds with no errors
- [x] Equivalence tests validate `forEach`, `sortIndex`, `flattenToElements`, and `flattenToContainers` produce identical results via reduce
- [x] Verify downstream snapshot tests still pass in CI

### DIFF
--- a/Example/UnitTests/AccessibilityHierarchyParserTests.swift
+++ b/Example/UnitTests/AccessibilityHierarchyParserTests.swift
@@ -204,23 +204,18 @@ final class AccessibilityHierarchyParserTests: XCTestCase {
         XCTAssertEqual(hierarchy.count, 1)
 
         // Verify it's a container with correct label
-        if case let .container(containerInfo, children) = hierarchy.first {
-            if case let .semanticGroup(label, _, _) = containerInfo.type {
-                XCTAssertEqual(label, "Group Label")
-            } else {
-                XCTFail("Expected semanticGroup container type")
-            }
-            XCTAssertEqual(children.count, 1)
-
-            // Verify child element
-            if case let .element(childElement, _) = children.first {
-                XCTAssertEqual(childElement.description, "Element")
-            } else {
-                XCTFail("Expected element child")
-            }
+        let containers = hierarchy.flattenToContainers()
+        XCTAssertEqual(containers.count, 1)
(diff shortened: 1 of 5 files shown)


soroushsq (commented):
Should we just conform to Collection here and then we'll get all these operations for free?

RoyalPineapple:
collection hadnt occured to me, great idea let me see what it takes.

===A few hours later===

Looked into this — unfortunately it doesn’t quite work out because these operations preserve the tree structure, they arent flat sequence operations. Sequence.filter returns [Element] which would destroy the tree structure, and Sequence.map wouldn’t give us bottom-up semantics (where the closure sees already-mapped children). reduce is the closest fit but conforming just for that one operation while keeping custom filter and map would be more confusing than helpful.

PR #328, opened 2026-04-05
Fix parser missing search fields behind zero-frame SwiftUI wrapper views

RoyalPineapple:
## Problem

SwiftUI `.searchable(text:prompt:placement:)` search bars are invisible to `AccessibilityHierarchyParser`. VoiceOver can reach the search field, but the parser never includes it in its output — so accessibility snapshots silently omit it.

### Root cause

On iOS 26+, UIKit renders the search bar through SwiftUI bridging views inside a `_UIFloatingBarContainerView`. The view hierarchy looks like this:

```
UILayoutContainerView
  ├─ UINavigationTransitionView
  ├─ UINavigationBar
  └─ _UIFloatingBarContainerView
       └─ FloatingBarHostingView
            └─ _UIInheritedView  ← frame {0, 0, 0, 0}  ⛔ parser stops here
                 └─ UIPlatformGlassInteractionView
                      └─ UICorePlatformViewHost<InlineSearchBarViewRepresentation>
                           └─ UISearchBarTextField  ← isAccessibilityElement, traits=searchField
```

The `_UIInheritedView` wrapper has a zero-size frame. The parser's visibility guard unconditionally pruned all zero-frame views, so the entire subtree — including the search field — was never traversed.

## Fix

```diff
- if let \`self\` = self as? UIView, self.isHidden || self.frame.size == .zero || self.alpha <= 0 {
+ if let \`self\` = self as? UIView,
+    self.isHidden || self.alpha <= 0
+    || (self.frame.size == .zero && (self.isAccessibilityElement || self.accessibilityElements != nil))
+ {
```

Zero-frame views are now only pruned when they are **accessibility elements** or **explicit accessibility containers** (`accessibilityElements != nil`). Plain wrapper views that only contain subviews are allowed through — their children can overflow with `clipsToBounds = false` (the UIKit default) and be fully visible on screen.

This preserves existing behavior for intentionally zero-sized accessibility containers (validated by `testZeroSizedContainerInElementStack`) while allowing the parser to traverse SwiftUI bridging wrappers.

## Before / After

| Before | After |
|--------|-------|
| Before | After |

**Before:** The search bar is visible on screen but completely absent from the accessibility snapshot legend — the parser only finds 2 elements.

**After:** The parser discovers the search field through the zero-frame wrapper. The legend now includes "Search: Search. Text Field. Search Field. *Double tap to edit.*" as expected.

## Changes

| File | What |
|------|------|
| `AccessibilityHierarchyParser.swift` | Narrow the zero-frame pruning guard (1 line changed) |
| `AccessibilityHierarchyParserTests.swift` | Add `testAccessibleChildrenFoundThroughZeroFrameNonClippingWrapper` |
| `SwiftUISearchableView.swift` | New SwiftUI demo view with `.searchable(placement: .toolbar)` |
| `SearchBarAccessibilityViewController.swift` | New UIKit demo with `UISearchController` |

## Test plan

- [x] All 49 existing unit tests pass (including `testZeroSizedContainerInElementStack`)
- [x] New unit test validates zero-frame wrapper traversal
- [x] Snapshot test target builds
- [x] Run full snapshot suite on CI to confirm no regressions

### DIFF
--- a/Example/AccessibilitySnapshot/SearchBarAccessibilityViewController.swift
+++ b/Example/AccessibilitySnapshot/SearchBarAccessibilityViewController.swift
@@ -1,0 +1,89 @@
+import UIKit
+
+/// Demonstrates a UINavigationController with a UISearchController attached, where the search
+/// bar is rendered through SwiftUI bridging views that include a zero-frame wrapper.
+final class SearchBarAccessibilityViewController: AccessibilityViewController {
+    // MARK: - Life Cycle
+
+    init() {
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - UIViewController
+
+    override func loadView() {
+        view = View()
+    }
+}
+
+// MARK: -
+
+extension SearchBarAccessibilityViewController {
+    final class ContentViewController: UIViewController {
+        override func viewDidLoad() {
+            super.viewDidLoad()
+            title = "Search Demo"
+            view.backgroundColor = .white
+
+            let searchController = UISearchController(searchResultsController: nil)
+            searchController.obscuresBackgroundDuringPresentation = false
+            navigationItem.searchController = searchController
+            navigationItem.hidesSearchBarWhenScrolling = false
+
+            let stackView = UIStackView()
+            stackView.axis = .vertical
+            stackView.spacing = 12
+            stackView.alignment = .center
+            stackView.translatesAutoresizingMaskIntoConstraints = false
+            view.addSubview(stackView)
+
+            for i in 1 ... 3 {
+                let label = UILabel()
+                label.text = "Item \(i)"
+                label.isAccessibilityElement = true
+                stackView.addArrangedSubview(label)
+            }
+
+            NSLayoutConstraint.activate([
+                stackView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+                stackView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor, constant: 20),
+            ])
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


RoyalPineapple:
before

Example/AccessibilitySnapshot/SearchBarAccessibilityViewController.swift
```diff
@@ -0,0 +1,89 @@
+import UIKit
+
+/// Demonstrates a UINavigationController with a UISearchController attached, where the search
+/// bar is rendered through SwiftUI bridging views that include a zero-frame wrapper.
+final class SearchBarAccessibilityViewController: AccessibilityViewController {
```
johnnewman-square, line 5: Should this be added to the RootViewController's `accessibilityScreens` array or be included in snapshot tests?

johnnewman-square (approved):
This looks good! Do you think we could add some new snapshot tests to this PR to showcase the enhanced behavior?

robmaceachern (commented):
One non-blocking concern: the new zero-frame visibility guard in AccessibilityHierarchyParser.swift is broader than the evidence in the tests currently proves. The implementation now traverses any zero-frame, non-clipping UIView subtree, but the added test only covers the specific search-bar happy path. I would prefer either tightening the predicate to the intended wrapper cases or adding a regression test that makes the broader contract explicit.

---

robmaceachern (approved):

RoyalPineapple (commented):
Good catch, thanks @johnnewman-square! Added `SearchBarAccessibilityViewController` to the `accessibilityScreens` array in RootViewController.

RoyalPineapple (commented):
Thanks for the feedback @robmaceachern! The predicate has been tightened (it now checks `clipsToBounds || isAccessibilityElement || accessibilityElements != nil`), and I've added a regression test (`testAccessibleChildrenPrunedBehindZeroFrameClippingWrapper`) that verifies zero-frame clipping wrappers are still pruned.

PR #335, opened 2026-05-19
Fall back to UIAccessibilityContainer index APIs when accessibilityElements is nil

RoyalPineapple:
## Summary
- When parsing an object whose `accessibilityElements` property is `nil`, fall back to the `accessibilityElementCount()` / `accessibilityElement(at:)` APIs before dropping through to subview traversal. This mirrors how VoiceOver resolves containers that vend elements via the dynamic `UIAccessibilityContainer` methods.
- A non-`nil` array is still honored as-is (including an empty array), matching Apple's documented semantics ("This can be used as an alternative to implementing the dynamic methods. default == nil").
- Logs a warning when a container reports more elements via `accessibilityElementCount()` than it actually returns from `accessibilityElement(at:)`.

## Test plan
- [ ] CI green
- [ ] Existing snapshot tests still pass

### DIFF
--- a/Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
+++ b/Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
@@ -782,13 +782,20 @@ private extension NSObject {
 
     /// Returns the explicit accessibility elements exposed by this object. When
     /// `accessibilityElements` is set (including to an empty array) it is used directly. When it is
-    /// `nil`, this falls back to the `accessibilityElementCount()` / `accessibilityElement(at:)`
-    /// container APIs, mirroring the order in which `UIAccessibilityContainer` consumers (including
-    /// VoiceOver) resolve elements. Returns `nil` when the object does not act as an explicit
-    /// accessibility container.
+    /// `nil` and the receiver is not a `UIView`, this falls back to the `accessibilityElementCount()` /
+    /// `accessibilityElement(at:)` container APIs, mirroring how `UIAccessibilityContainer` consumers
+    /// (including VoiceOver) resolve elements. For `UIView` instances the existing subview-iteration
+    /// path is used instead — UIKit synthesizes counts from the subview tree when flags like
+    /// `shouldGroupAccessibilityChildren` are set, which would otherwise duplicate or recurse on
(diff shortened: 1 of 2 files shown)


Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -778,6 +780,45 @@ private extension NSObject {
...
+        if let elements = accessibilityElements as? [NSObject] {
+            return elements
+        }
+
+        guard !(self is UIView) else {
+            return nil
+        }
+
+        let count = accessibilityElementCount()
+        guard count > 0 else {
```
RoyalPineapple: **🔴 Infinite-loop / CI hang: `count > 0` doesn't exclude `NSNotFound`.**

`accessibilityElementCount()`'s default return is undocumented in `UIAccessibilityContainer.h` (its siblings `accessibilityElementAtIndex:` → `nil` and `indexOfAccessibilityElement:` → `NSNotFound` *are* documented). On **iOS 17/18/26 it returns `NSNotFound`** (`9_223_372_036_854_775_807`) for objects that don't implement the dynamic container methods. That passes `count > 0`, so `for index in 0 ..< count` below runs ~9.2 quintillion iterations and hangs.

This is what's failing CI: all three Tuist Build jobs ran 6h0m and were canceled at the timeout. The iOS_18 log executes normally through `testLargeView` then goes silent for ~5h46m, hanging on the next test (`testLargeViewInViewControllerThatRequiresTiling`, a VC-hierarchy case where the parser reaches UIKit-internal non-UIView objects). (Note: on iOS 16.2 this method returns `0`, so it won't reproduce on older local runtimes.)

Fix:
```suggestion
        let count = accessibilityElementCount()
        guard count > 0, count != NSNotFound else {
```
After this, `testLargeViewInViewControllerThatRequiresTiling` provides regression coverage for free.

Minor (separate line): the warning below uses `print(...)`; the rest of this file uses `os_log(..., log: parserLog, type: .error, ...)` (see `providedContextAsSuperview`). Worth matching for consistency / to avoid stdout spam in host test processes.

RoyalPineapple:
🤖 Superseded by #344 which now targets main and includes both the fallback and the guard removal.

PR #336, opened 2026-05-20
Scope viewToElementsMap to parse call lifetime

RoyalPineapple:
## Summary

`viewToElementsMap` is currently scoped to the `AccessibilityHierarchyParser` instance. This means every tab bar view and its child `NSObject`s parsed across calls are retained for the lifetime of the parser. When the parser is long-lived, this is a memory leak.

This PR narrows the scope from instance to function. The dictionary is declared as a local in `parseAccessibilityHierarchy(in:)` and passed `inout` to `context(for:from:...)`. Same caching behavior within a single parse call, no retention across calls.

## Changes

- Remove `viewToElementsMap` instance property  
- Declare as local in `parseAccessibilityHierarchy(in:)`, pass `inout` to `context(for:from:...)`

## Test plan

- Existing parser tests pass unchanged — cache behavior within a single parse call is identical

### DIFF
--- a/Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
+++ b/Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
@@ -160,6 +160,7 @@ public final class AccessibilityHierarchyParser {
             userInterfaceIdiom: userInterfaceIdiom
         )
 
+        var viewToElementsMap: [UIView: [NSObject]] = [:]
         let contextualizedElements = uncontextualizedElements.map { element in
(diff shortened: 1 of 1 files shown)


Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -487,9 +490,6 @@ public final class AccessibilityHierarchyParser {
 
     // MARK: - Private Properties
```
RoyalPineapple:
```suggestion
```

soroushsq (approved):
I like it

PR #344, opened 2026-06-04
Fall back to UIAccessibilityContainer index APIs when accessibilityElements is nil

RoyalPineapple:
## Summary

Falls back to the `UIAccessibilityContainer` index APIs (`accessibilityElementCount()` / `accessibilityElement(at:)`) when `accessibilityElements` returns nil.

Some containers vend their children only through the index-based container APIs, returning nil from `accessibilityElements`. The parser previously treated these as having no explicit elements and fell through to subview traversal, missing the container-vended children. This change mirrors the behavior of UIKit's own accessibility tree walker, which uses the index APIs for any object that reports a valid count — including UIViews.

## How it works

`resolvedAccessibilityElements()` checks `accessibilityElements` first. If nil, it iterates `accessibilityElement(at:)` up to `accessibilityElementCount()`. Objects that report `NSNotFound` from `accessibilityElementCount()` (the default) fall through unchanged — their existing traversal path (subview iteration with grouping, sorting, and pruning) is preserved.

In practice, no standard UIKit UIView uses the index API without also setting `accessibilityElements` — UITableView, UICollectionView, UIScrollView, etc. all report either `NSNotFound` or count=0 with nil elements. The fallback primarily captures non-UIView containers (e.g. `UIAccessibilityElement` subclasses) that implement the dynamic container methods.

## Validation

Validated against `_accessibilityLeafDescendantsWithOptions:` (UIKit's internal tree walker) with default options, VoiceOver options, and scanner groups mode. All three confirm the index API takes precedence over subview traversal when implemented.

## Changes

- Add `resolvedAccessibilityElements()` on `NSObject` — tries `accessibilityElements`, falls back to index APIs, filters `NSNotFound` and non-positive count sentinels
- Tests for index-API fallback (both UIView and non-UIView), `NSNotFound` sentinel, zero-count, `UISegmentedControl` end-to-end, and `UIView` container resolution

### DIFF
--- a/Example/UnitTests/AccessibilityHierarchyParserTests.swift
+++ b/Example/UnitTests/AccessibilityHierarchyParserTests.swift
@@ -993,6 +993,126 @@ final class AccessibilityHierarchyParserTests: XCTestCase {
         XCTAssertNotNil(listNode.container)
     }
 
+    // MARK: - UIAccessibilityContainer Index API Fallback
+
+    /// A non-`UIView` `NSObject` that vends elements via the index-based container APIs
+    /// (`accessibilityElementCount()` / `accessibilityElement(at:)`) while leaving
+    /// `accessibilityElements` nil should have its elements resolved through that fallback.
+    func testNonViewContainerResolvesElementsViaIndexAPIs() {
+        let rootView = UIView(frame: .init(x: 0, y: 0, width: 200, height: 100))
+
+        let child1 = UIAccessibilityElement(accessibilityContainer: rootView)
+        child1.accessibilityLabel = "Index Child 1"
+        child1.accessibilityFrame = CGRect(x: 0, y: 0, width: 200, height: 40)
+
+        let child2 = UIAccessibilityElement(accessibilityContainer: rootView)
+        child2.accessibilityLabel = "Index Child 2"
+        child2.accessibilityFrame = CGRect(x: 0, y: 50, width: 200, height: 40)
+
+        // The container leaves `accessibilityElements` nil and only implements the index APIs.
+        let container = IndexAPIContainer(elements: [child1, child2])
+        rootView.accessibilityElements = [container]
+
+        let parser = AccessibilityHierarchyParser()
+        let elements = parser.parseAccessibilityHierarchy(
+            in: rootView,
+            userInterfaceLayoutDirectionProvider: TestUserInterfaceLayoutDirectionProvider(userInterfaceLayoutDirection: .leftToRight),
+            userInterfaceIdiomProvider: TestUserInterfaceIdiomProvider(userInterfaceIdiom: .phone)
+        ).flattenToElements().map { $0.description }
+
+        XCTAssertEqual(elements, ["Index Child 1", "Index Child 2"])
+    }
+
+    /// `accessibilityElementCount()` returns `NSNotFound` (== `NSIntegerMax`) for objects that
+    /// don't actually implement the dynamic container methods. The parser must treat that sentinel
+    /// as "no elements" rather than iterating ~`NSIntegerMax` times. This test would hang if the
+    /// guard were removed, so it doubles as a regression test for the sentinel handling.
+    func testNonViewContainerWithNSNotFoundCountIsTreatedAsEmpty() {
+        let rootView = UIView(frame: .init(x: 0, y: 0, width: 200, height: 100))
+
+        let child = UIAccessibilityElement(accessibilityContainer: rootView)
+        child.accessibilityLabel = "Should Not Appear"
+        child.accessibilityFrame = CGRect(x: 0, y: 0, width: 200, height: 40)
+
+        // Report the `NSNotFound` sentinel even though an element technically exists.
+        let container = IndexAPIContainer(elements: [child], reportedCount: NSNotFound)
+        rootView.accessibilityElements = [container]
+
+        let parser = AccessibilityHierarchyParser()
+        let elements = parser.parseAccessibilityHierarchy(
+            in: rootView,
+            userInterfaceLayoutDirectionProvider: TestUserInterfaceLayoutDirectionProvider(userInterfaceLayoutDirection: .leftToRight),
+            userInterfaceIdiomProvider: TestUserInterfaceIdiomProvider(userInterfaceIdiom: .phone)
+        ).flattenToElements().map { $0.description }
+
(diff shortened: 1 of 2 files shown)


Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -882,6 +884,45 @@ private extension NSObject {
...
+        elements.reserveCapacity(count)
+        for index in 0 ..< count {
+            // `accessibilityElement(at:)` may return `nil` (or a non-`NSObject`) for individual indices even
+            // when `accessibilityElementCount()` reports a larger count; those are skipped rather than treated
+            // as a hard error, mirroring how a consumer iterating the container would tolerate gaps.
+            if let element = accessibilityElement(at: index) as? NSObject {
+                elements.append(element)
+            }
+        }
+        return elements.isEmpty ? nil : elements
```
robmaceachern, line 923: blocker: Preserve explicit-container precedence when all indexed slots are invalid

A positive count selects the dynamic container path, but filtering every slot to an empty array turns it back into `nil`. For a `UIView`, that switches to subview traversal at line 854, so native subviews leak into the snapshot even though the index APIs should take precedence; it also differs from an explicit `accessibilityElements = []`. Return the possibly empty `elements` array after a valid positive count, and cover `count = 1` with a nil slot plus an accessible real subview to lock this precedence.

---

Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -819,9 +819,11 @@ private extension NSObject {
         // _UIInheritedView inside _UIFloatingBarContainerView) use zero-frame wrappers whose children overflow and
         // are visible. Pruning those hides real accessible content such as the UISearchBarTextField rendered by
         // .searchable().
+        let explicitAccessibilityElements = resolvedAccessibilityElements()
```
robmaceachern, line 822: suggestion: Prune hidden and transparent views before resolving dynamic elements

This call now runs before the existing `isHidden` / alpha short-circuit. A hidden custom view with nil `accessibilityElements` and a finite count will enumerate every index even though line 824 immediately discards it; the base implementation never evaluated its elements because the visibility condition short-circuited first. Split the visibility/alpha guard ahead of resolution, leaving the zero-frame decision after it, and add a call-count regression so hidden reusable views cannot introduce unexpected work or side effects.

---

Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -882,6 +884,45 @@ private extension NSObject {
...
+        // `accessibilityElement(at:)` and `index(ofAccessibilityElement:)` document `nil` / `NSNotFound`
+        // defaults, and `NSNotFound` is what UIKit returns on iOS 17+). Treating that sentinel as a real
+        // count would iterate ~`NSIntegerMax` times, so guard against it explicitly.
+        let count = accessibilityElementCount()
+        guard count > 0, count != NSNotFound else {
+            return nil
+        }
+
+        var elements: [NSObject] = []
+        elements.reserveCapacity(count)
```
robmaceachern, line 914: suggestion: Bound implausible positive counts before reserving and iterating

`NSNotFound` is only one invalid or stale value. A container returning `Int.max - 1`, or another implausibly large finite count, still reaches `reserveCapacity` and can terminate the host or spend unbounded time dispatching `accessibilityElement(at:)`. Since this parser runs in the client process and already treats the sentinel defensively, reject and log an implementation-defined safety limit before allocation/iteration. A fake container can cover this by asserting that indexed lookup is never called.

---

Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -882,6 +884,45 @@ private extension NSObject {
...
+        for index in 0 ..< count {
+            // `accessibilityElement(at:)` may return `nil` (or a non-`NSObject`) for individual indices even
+            // when `accessibilityElementCount()` reports a larger count; those are skipped rather than treated
+            // as a hard error, mirroring how a consumer iterating the container would tolerate gaps.
+            if let element = accessibilityElement(at: index) as? NSObject {
```
robmaceachern, line 919: blocker: Detect cycles on the active traversal path

An index container can vend itself or an ancestor; once appended here, line 840 recursively enters that object without termination. This is newly reachable for dynamic-only containers, which were not traversed while `accessibilityElements` was nil before this change. Thread active ancestor identities through `recursiveAccessibilityHierarchy` and skip/log back-edges only (rather than globally deduplicating, so legitimate repeated elements remain). Please cover both a self-reference and a two-container cycle.

---

Example/UnitTests/AccessibilityHierarchyParserTests.swift
```diff
@@ -993,6 +993,126 @@ final class AccessibilityHierarchyParserTests: XCTestCase {
...
+        let elements = parser.parseAccessibilityHierarchy(
+            in: rootView,
+            userInterfaceLayoutDirectionProvider: TestUserInterfaceLayoutDirectionProvider(userInterfaceLayoutDirection: .leftToRight),
+            userInterfaceIdiomProvider: TestUserInterfaceIdiomProvider(userInterfaceIdiom: .phone)
+        ).flattenToElements().map { $0.description }
+
+        XCTAssertTrue(elements.isEmpty, "Expected the NSNotFound count to be treated as no elements, got: \(elements)")
+    }
+
+    func testContainerWithZeroCountIsTreatedAsEmpty() {
```
robmaceachern, line 1051: suggestion: Make the zero-count test distinguish the new traversal behavior

This test also passes on the base revision: a non-`UIView` object with no explicit elements produced no nodes before the fallback. To guard the intended compatibility branch, use an index-backed `UIView` reporting zero with an accessible subview and assert that the subview is still traversed. Otherwise `resolvedAccessibilityElements()` could incorrectly return `[]` for zero and this test would remain green.

---

robmaceachern (approved):
The bot generated some feedback that might be worth considering but lgtm.

PR #347, opened 2026-07-10
Suppress derived user-input-label echo in the parser

RoyalPineapple:
## Summary

When an element has no explicitly-set Voice Control input labels, UIKit synthesizes a fallback of `[accessibilityLabel]`. That echo is redundant for targeting — you can already invoke the element by speaking its label — and is byte-identical on the wire to setting nothing. The parser was surfacing this derived echo as if it were an authored input label, producing a redundant input-label "pill" in the legend for any element whose only accessibility property is its label (e.g. every plain `UITableViewCell`).

## Fix

Add `NSObject.authoredUserInputLabels`, which returns `nil` for the single-element `[accessibilityLabel]` echo and otherwise passes labels through, and use it in `buildElement`.

Note: the private raw attributed getter (`_accessibilityAXAttributedUserInputLabels`) is *not* a reliable discriminator — some classes (notably `UITableViewCell` via its axbundle) surface the echo through it too. The only dependable signal that input labels are a genuine authorial override is that they differ from the accessibility label, which is what this checks.

## Tests

- `testDerivedUserInputLabelEchoIsSuppressed` — a label-only element yields `userInputLabels == nil`.
- `testAuthoredUserInputLabelsArePreserved` — distinct authored input labels (`["Send", "Go"]`) are kept.

## Evidence

Grounded in decompilation of iOS 26.x UIAccessibility: the `[label]` echo is synthesized in `_accessibilityAXAttributedUserInputLabelsIncludingFallbacks` as a fallback when the raw value is empty, and Voice Control's fetch path resolves the same fallback — so `userInputLabels == [label]` is indistinguishable from "unset" for targeting purposes.


Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
```diff
@@ -260,7 +260,7 @@ public final class AccessibilityHierarchyParser {
             traits: AccessibilityTraits(object.accessibilityTraits),
             identifier: object.identifier,
             hint: hint,
-            userInputLabels: object.accessibilityUserInputLabels,
+            userInputLabels: object.authoredUserInputLabels,
```
johnnewman-square, line 263: 🤖 This normalization also affects the SwiftUI renderer’s `.always` mode. UIKit rebuilds default labels when `userInputLabels` is nil, but the SwiftUI renderer only displays non-empty marker labels. For affected cells, `.always` therefore changes from showing the derived label to showing no input-label pills. Could we pass the full display mode through SwiftUI and synthesize defaults there as UIKit does, with a SwiftUI `.always` regression test?

Example/UnitTests/AccessibilityHierarchyParserTests.swift
```diff
@@ -993,6 +993,40 @@ final class AccessibilityHierarchyParserTests: XCTestCase {
...
+    // MARK: - User Input Labels
+
+    /// An element whose only accessibility property is its label reports a Voice Control input-label
+    /// echo (`[accessibilityLabel]`) that UIKit synthesizes as a fallback. That echo is redundant for
+    /// targeting (you can already invoke the element by speaking its label), so the parser must not
+    /// surface it as an authored input label.
+    func testDerivedUserInputLabelEchoIsSuppressed() {
+        let container = UIView(frame: .init(x: 0, y: 0, width: 100, height: 100))
+        let element = UIView(frame: .init(x: 0, y: 0, width: 100, height: 44))
+        element.isAccessibilityElement = true
```
johnnewman-square, line 1007: 🤖 On iOS 26.5, this plain `UIView` returns `[]` from `accessibilityUserInputLabels`, not `["Submit"]`, so this test only exercises the empty-array-to-nil normalization. I confirmed that removing the `[accessibilityLabel]` comparison still leaves both added tests passing. Could this use a deterministic subclass returning `[accessibilityLabel]`—and ideally an accessible `UITableViewCell` case—and assert that precondition before parsing?

johnnewman-square (approved):
This looks good to me. My agent flagged a few issues above.

Example/UnitTests/AccessibilityHierarchyParserTests.swift
RoyalPineapple: 🤖 Good catch — on iOS 26 the plain-view getter indeed returns `[]`, so the test was only pinning the empty-to-nil normalization there. Fixed in dcc024fe: the test now uses an `InputLabelEchoingView` subclass whose getter deterministically returns `[accessibilityLabel]`, and asserts that precondition before parsing. I deliberately skipped the `UITableViewCell` variant: whether the cell's axbundle surfaces the echo is exactly the OS-dependent behavior this comment flags, and the subclass pins the same parser branch deterministically on every OS version.

Sources/AccessibilitySnapshot/Parser/Swift/Classes/AccessibilityHierarchyParser.swift
RoyalPineapple: 🤖 Fixed in dcc024fe by sharing the resolution logic instead of duplicating it: `AccessibilityMarker.displayUserInputLabels(_:)` in Core now backs both legend renderers, and the SwiftUI legend takes the full `AccessibilityContentDisplayMode` instead of a flattened Bool. `.always` synthesizes the Voice Control defaults (label split on spaces, plus trait phrases) and `.whenOverridden` now also gates on `respondsToUserInteraction`, matching UIKit. Added deterministic unit tests for the helper plus a SwiftUI `.always` snapshot regression test (`testBasicAccessibilityDemoWithInputLabelsAlways`). Note: the iOS 18.5 reference is recorded; the 26.2 reference still needs recording (no 26.2 runtime on this machine) — it can be pulled from the CI xcresult artifacts on the first run.

Sources/AccessibilitySnapshot/AccessibilitySnapshotPreviews/SwiftUILegendView.swift
```diff
@@ -7,18 +7,18 @@ import SwiftUI
...
     public let palette: ColorPalette
-    public let showUserInputLabels: Bool
+    public let inputLabelDisplayMode: AccessibilityContentDisplayMode
     public let showUnspokenTraits: Bool
 
     public init(
         markers: [AccessibilityMarker],
         palette: ColorPalette,
-        showUserInputLabels: Bool,
+        inputLabelDisplayMode: AccessibilityContentDisplayMode,
```
robmaceachern, line 16: blocker: Preserve the existing public `LegendView` API while adding the enum-based mode

`LegendView` is exported by the `AccessibilitySnapshotPreviews` product. Replacing the public `showUserInputLabels:` initializer argument and stored property with `inputLabelDisplayMode:` makes existing direct initializer calls and property reads stop compiling. The enum is a better long-term shape, but this parser fix does not need to impose a source-breaking migration.

Please keep the new enum-based initializer as the preferred API while adding a deprecated `showUserInputLabels:` overload and compatibility property that delegate to the new representation. Alternatively, keep the Boolean initializer public and make the enum initializer internal if only package renderers need the full mode.

---

robmaceachern (approved):
lgtm. agent flagged a backwards compat question

Issue #75, opened 2021-09-29
Large views silently fail to render using drawHierarchyInRect rendering mode


NickEntin:
Views that are very large in at least one dimension (reproduced with a view that was 2870 pt tall) will silently fail to render when using the `.drawHierarchyInRect` rendering mode, producing an empty snapshot of the view.

cc @efirestone

NickEntin:
This problem is actually happening at two separate levels: when we snapshot the view under test _and_ when we snapshot the container view. In some cases, where we're right at the threshold, we may successfully snapshot the view under test but fail to snapshot the container view (since the container view is larger in most cases).

I consider this to be three separate bugs: a bug in our rendering of the view under test, and a bug in each of the snapshotting engines we use for snapshotting the container view. I opened #76 to address the rendering of the view under test. We should probably open issues against iOSSnapshotTestCase and SnapshotTesting to apply similar fixes.

There's sort of a workaround we could apply when using SnapshotTesting - we could always use `drawHierarchyInKeyWindow: false` when snapshotting the container view. In _theory_ there shouldn't be any difference since the container view doesn't have a real reason to use that rendering mode (there aren't any 3D transforms, effects views, etc.); however, from testing there are actually some minor rendering differences, which means consumers would need to upload all of their reference images. This isn't ideal, so I think fixing the issue inside SnapshotTesting is preferable.

NickEntin:
I'm marking this as `blocked` since we've resolved as much of it as we reasonably can within AccessibilitySnapshot for now.

NickEntin:
I'm gonna go ahead and mark this as closed since there's no additional changes needed inside AccessibilitySnapshot to support this.

PR #352, opened 2026-08-11
Fix stale safe area in tiled hierarchy snapshots

johnnewman-square:
## Summary

Fixes misaligned accessibility overlays for views large enough to require tiled rendering (over 2,000pt in either dimension).

Tiled rendering temporarily restructures the view hierarchy around the snapshotted view. UIKit zeroes a view's safe area when it leaves its superview and doesn't recompute it in the new hierarchy until something invalidates it, so safe-area-driven layout (navigation bars, headers, bottom bars) renders shifted up by the safe area height. The accessibility markers are parsed afterwards from the restored hierarchy, so they remain correct — producing an annotated snapshot whose overlays don't line up with the screenshot underneath.

The existing compensation (introduced in #80) measures the safe area immediately after reparenting — reading that stale zeroed value — and applies the difference to a temporary view controller that sits outside the view's safe-area authority chain. A view owned by its own view controller only honors that controller's `additionalSafeAreaInsets`, so the compensation has no effect for view-controller-owned views, and is unreliable even for plain views.

## Fix

Restore the safe area through the view controller that actually governs the view: its owning view controller when it has one, or a temporary view controller that adopts the view otherwise. Toggling that controller's `additionalSafeAreaInsets` (set +1pt, restore — both before the next layout pass, so the transient value is never rendered) forces UIKit to recompute the stale safe area, which resolves back to the original value on its own. If the original value can't be restored, rendering now fails with a new `ImageRenderingError.containedViewSafeAreaCouldNotBePreserved` instead of silently producing shifted content.

Views that never enter the tiled path are untouched. Tiled views whose layout doesn't consume the safe area render identically to before.

## Testing

The existing tiling snapshot tests are disabled due to #75, so coverage comes in two parts:

- **Unit tests** (`ViewTiledImageRenderingTests`) that assert on rendered pixels: stripes laid out relative to the top safe area inset must appear in the tiled image at the same rows they occupy in the original hierarchy — one inside the first tile, one beyond the first tile boundary — plus post-render restoration checks. Covered shapes: a view owned by a view controller, a plain un-owned view, and a view controller with `additionalSafeAreaInsets` set (the scenario from #80). All three fail against the previous implementation and pass with this change. The full UnitTests bundle passes.
- **A snapshot test** (`testTallViewInViewControllerThatRequiresTiling`) exercising the full annotated-snapshot pipeline on a 3,280pt-tall view-controller-owned view whose content is anchored to its top safe area inset. It sidesteps #75 by passing `.drawHierarchyInRect` only through `snapshotConfiguration` (the inner tiled render under test) while the outer snapshot comparison keeps the default layer rendering. Against the previous implementation, the recorded reference shows the screenshot shifted up by the device's safe area height with detached overlays; with this change, content and overlays align. Reference images are recorded for all three CI device configurations (iOS 17.5, 18.5, and 26.2).


Sources/AccessibilitySnapshot/Core/UIView+ImageRendering.swift
```diff
@@ -145,24 +184,48 @@ extension UIView {
...
-            top: originalSafeArea.minY - currentSafeArea.minY,
-            left: originalSafeArea.minX - currentSafeArea.minX,
-            bottom: currentSafeArea.maxY - originalSafeArea.maxY,
-            right: currentSafeArea.maxX - originalSafeArea.maxX
-        )
+
+        // Reparenting the view can leave its safe area stale: UIKit zeroes the safe area when the view is removed
+        // from its superview and does not recompute it in the new hierarchy until something invalidates it. Toggling
+        // the governing view controller's additional safe area insets forces that recomputation, after which the safe
+        // area generally resolves to its original value on its own.
```
RoyalPineapple: It looks like we've got two mechanisms at play here which seem redundant and are both manipulating the same `additionalSafeAreaInsets` 

> 
> Mechanism 1 — the nudge (provoke UIKit to fix itself):
> ```
>   var nudgedSafeAreaInsets = originalAdditionalSafeAreaInsets
>   nudgedSafeAreaInsets.top += 1
>   governingViewController.additionalSafeAreaInsets = nudgedSafeAreaInsets          // poke
>   governingViewController.additionalSafeAreaInsets = originalAdditionalSafeAreaInsets  // restore
>   RunLoop.current.run(until: Date())
>  ```
>   Mechanism 2 — the delta compensation (force it mathematically if the poke wasn't enough):

> ```
>   let recomputedSafeAreaInsets = safeAreaInsets
>   if recomputedSafeAreaInsets != originalSafeAreaInsets {
>       governingViewController.additionalSafeAreaInsets = UIEdgeInsets(
>           top:    originalAdditionalSafeAreaInsets.top    + originalSafeAreaInsets.top    - recomputedSafeAreaInsets.top,
>           left:   originalAdditionalSafeAreaInsets.left   + originalSafeAreaInsets.left   - recomputedSafeAreaInsets.left,
>           bottom: originalAdditionalSafeAreaInsets.bottom + originalSafeAreaInsets.bottom - recomputedSafeAreaInsets.bottom,
>           right:  originalAdditionalSafeAreaInsets.right  + originalSafeAreaInsets.right  - recomputedSafeAreaInsets.right
>       )
>       RunLoop.current.run(until: Date())
>   }
> ```
> 
> 



have you tried running it with mechanism 2 alone? now that its being applied to the correct owning view controller I suspect the bug doesnt need an additional nudge.

RoyalPineapple (approved):
LTGM one question about the belt-and-supspenders aspect of the fix

RoyalPineapple (approved):

Sources/AccessibilitySnapshot/Core/UIView+ImageRendering.swift
johnnewman-square: Thanks for the suggestion @RoyalPineapple. The first mechanism to nudge the safe area is what we needed to fix the layout behavior. I've pushed up a simplification and updated the description.

Here's a full agentic report:
> 🤖 I tested both directions, and it turns out the redundancy runs the other way: the nudge is the load-bearing part, and mechanism 2 can't stand alone.
>
> Mechanism 2 alone (nudge removed): all three regression tests fail with the safe area over-inflated by exactly the window-propagated insets — top 59 → 118, bottom 34 → 68, and 159 → 218 in the additionalSafeAreaInsets test. The trap is the same stale-cache behavior that broke the original #80 compensation: the delta is computed from the stale post-reparent read, but writing additionalSafeAreaInsets is itself what invalidates the cache — so UIKit recomputes the genuine base underneath the delta that was just applied, and the base gets counted twice. Any "compensate from a pre-invalidation read" approach self-defeats this way; making it converge would mean write → re-read → iterate, which is just the nudge with extra steps.
>
> Mechanism 1 alone (compensation removed): all tests pass, and the tall-view snapshot test renders pixel-identical to its reference. After the toggle forces recomputation, the safe area resolves to the original value on its own in every environment I've tested (iPhone and iPad, plain and VC-owned views).
>
> Since the compensation never fires in practice and can't replace the nudge, I've removed it in 2b75690 — the toggle plus the loud containedViewSafeAreaCouldNotBePreserved guard is the whole mechanism now. If some environment ever resolves to different insets after recomputation, rendering fails with both values in the error rather than silently producing a shifted snapshot.

PR #364, opened 2026-09-17
Don't reserve legend space for SwiftUI snapshots with no accessibility elements

RoyalPineapple:
Two independent causes made zero-element snapshots much larger than their content under the SwiftUI layout engine.

`PreParsedAccessibilitySnapshotView.contentWidth` floored the width at `LegendLayoutMetrics.minimumWidth` (284 + 16×2 = 316pt) unconditionally, so a 100pt-wide view rendered into a 316pt frame padded with grey. `body` now renders `snapshotWithOverlays` alone when `markers.isEmpty` — it is already framed to `renderSize`.

`LegendView` applied `.padding(legendInset)` around its `VStack` regardless of content, so an empty legend still occupied a 32pt band. It now returns `EmptyView()` when there are no markers, which also covers the live-preview path in `AccessibilitySnapshotView` that builds its own `LegendView`.

Neither branch changes layout when markers exist, so only zero-element snapshots should move.


Sources/AccessibilitySnapshot/AccessibilitySnapshotPreviews/SwiftUIAccessibilitySnapshotView.swift
```diff
@@ -206,7 +206,11 @@ public struct PreParsedAccessibilitySnapshotView: View {
     }
 
     public var body: some View {
-        if legendOnRight {
+        if markers.isEmpty {
+            // Without a legend there is nothing to reserve `minimumWidth` for.
+            snapshotWithOverlays
+                .background(Color(white: 0.9))
```
RoyalPineapple: this color should be defined and named somewhere

robmaceachern (approved):

